PR #14, opened 2014-08-13
add link in community packages section to mono-opt packages

tpokorra:
it might be helpful for people to get their hands on a 3.6.0 build of Mono for their distribution


akoeplinger:
I think this is a good idea until the Xamarin-provided packages are ready. Thanks!

PR #19, opened 2014-08-15
Fixed GTK# link in its first paragraph

lextm:
The original beginning part is a broken link whose syntax is unknown. Fixed it and linked to a valid GTK# page.


akoeplinger:
Thanks!

PR #22, opened 2014-08-16
Added a link to downstream packages

Mailaender:
To display which version is available where already instead of just the message that @xamarin is working on something on their own which I doubt will be there in time. Added a new section and sorted the distributions alphabetically.


akoeplinger:
This looks like a good idea until the xamarin packages are ready.

@directhex do you think we should keep this distinction between Xamarin and distribution packages in the future too? Or do we want to encourage using the Xamarin packages (once they are available)?

PR #26, opened 2014-08-19
missing dependency 'gettext'

TheAngryByrd:
I had a missing dependency of 'gettext' when attempting to run the script on my Debian 7.6 distro.


akoeplinger:
Thanks!

txdv:
Well, yeah, I tested it on Ubuntu Desktop so that thing has automatically gettext installed.

PR #27, opened 2014-08-28
Include C# 4.5 as feature complete.

misaelnieto:
reference: http://www.mono-project.com/docs/about-mono/languages/csharp/

### DIFF
--- a/docs/about-mono/index.md
+++ b/docs/about-mono/index.md
@@ -17,7 +17,7 @@ The Components
 
 There are several components that make up Mono:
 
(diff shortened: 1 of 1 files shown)


akoeplinger:
I think you're confusing the language C# with the .NET Framework. There is no C# language specification v4.5, only 1.0 - 5.0.

txdv:
http://en.wikipedia.org/wiki/C_Sharp_(programming_language)#Versions

close it

txdv:
wow, that table just made me realize that 3.0 4.0 and 5.0 have no official ECMA specifications.

that is exactly what all the mono haters said is going to happen

akoeplinger:
@txdv http://tirania.org/blog/archive/2014/Apr-09.html:

> For the last couple of months, the ECMA C# committee has been working on updating the spec to reflect C# 5.

PR #34, opened 2014-09-06
Fixed spelling

Mailaender:
> Like most names, it should be written in lower case with an initial capital, that is, "Lua". Please do not write it as "LUA", which is both ugly and confusing, because then it becomes an acronym with different meanings for different people. So, please, write "Lua" right!

### DIFF
--- a/docs/about-mono/languages/index.md
+++ b/docs/about-mono/languages/index.md
@@ -85,7 +85,7 @@ Object Pascal
 
 Their [Delphi Prism](http://www.codegear.com/products/delphi/prism) compiler support Mono out of the box.
 
-LUA
+Lua
 ---
 
 [Lua2Il](http://www.lua.inf.puc-rio.br/luanet/lua2il/) is a compiler that will allow you to run your existing Lua code or reuse the existent expertise you have on Lua in your application and run it with the Mono JIT compiler.
--- a/docs/about-mono/languages/index.md
+++ b/docs/about-mono/languages/index.md
@@ -88,7 +88,7 @@ Their [Delphi Prism](http://www.codegear.com/products/delphi/prism) compiler sup
 Lua
 ---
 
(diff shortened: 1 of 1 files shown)


docs/about-mono/languages/index.md
```diff
@@ -85,7 +85,7 @@ Object Pascal
 
 Their [Delphi Prism](http://www.codegear.com/products/delphi/prism) compiler support Mono out of the box.
 
-LUA
+Lua
 ---
 
 [Lua2Il](http://www.lua.inf.puc-rio.br/luanet/lua2il/) is a compiler that will allow you to run your existing Lua code or reuse the existent expertise you have on Lua in your application and run it with the Mono JIT compiler.
```
akoeplinger: The link doesn't work anymore, do you know if Lua2IL still exists? If not, maybe it's better to remove it here completely.

Mailaender: Fixed the link. Never worked with it though.

Mailaender: Pinging @mascarenhas

txdv:
LUA is such a nice language

Mailaender:
> ~~LUA~~ is such a nice language
> Lua is such a nice language

SCNR

akoeplinger:
I'm going to merge this, if we're hearing back from @mascarenhas that the project is dead we can always remove it then :smile:

txdv:
The link is kinda dead, just remove it.

mascarenhas:
Lua2IL is a dead project, but there are several active Lua bindings and implementations by @NLua. You might want to link to them. :-)

akoeplinger:
@mascarenhas Thanks for the update, added it in f7f2c8c2ea7b657851bffdd3bffe5d43c5dbce64.

PR #40, opened 2014-09-17
Update label.md

OnlineCop:
Converted spaces within inline code block to "&nbsp;" so formatting on webpage appears correctly.


Therzok:
Thanks!

PR #42, opened 2014-09-18
Another attempt at clarity on MDK download prompt

tbenade:


txdv:
FOR the Mono Runtime, by the way, you don't have close the pull request and issue a new one.

You can squash commits and force push

tbenade:
Sorry I edited _for_ as it sounds weird to me like the Mono Runtime is a cult or tribe of some sort. Example you don't develop for a house, you develop a house. At the end of the day you are the maintainer as you wish.

Thanks for the git tip, but I use GitHub as my IDE so I have no idea how to perform all that wizardry you just mentioned :grinning:

akoeplinger:
I think you're misunderstanding something here.
The MDK is not for developing Mono itself, it's really intended to be used by developers (as opposed to the MRE that is geared towards "normal" users that just want to run a .NET/Mono based application).

I think Matts answer to the [StackOverflow question](http://stackoverflow.com/questions/24868209/different-between-mre-and-mdk-installer) you link describes it pretty well, so maybe adapt something from that?

I'm all for clarifying the intent of MRE/MDK, but we should make sure it's not getting worse :smile:

tbenade:
Duh! I get it now sorry. I now understand the desire for "for" in explanation.

akoeplinger:
Can we add that the MDK is actually "MRE + Extra tools, libraries, .NET PCL (Portable Class Library) profiles, etc" to make it super clear?

akoeplinger:
@tbenade I pushed a slightly modified version and moved the explanation section below the download buttons, as it just looks better. Thanks for your help on this!

PR #43, opened 2014-09-23
fix code typo

mcgwiz:


akoeplinger:
Actually, gmcs isn't used anymore (see http://www.mono-project.com/docs/about-mono/languages/csharp/)

Can you change it to use mcs instead?

PR #45, opened 2014-09-24
Added SUSE Factory repo link

nilsh:
SUSE Factory updates quickly to the latest mono builds and provides a convenient way to install them; either via adding repository, with 1-click-install, or downloading the rpms.


akoeplinger:
Can you give me a quick overview of why this adds value / what's different to the other package link we already have? The other link by @tpokorra seems to have openSUSE packages as well.

nilsh:
Linked packages are built by the SUSE guys
- they are quick about updates
- this is a more common repository, also dependency-wise better integrated with other packages
- the the one already provided is more a private repository

akoeplinger:
Ok, thanks!

PR #48, opened 2014-09-27
Add fix for "No Application Found Unable to find a matching application ...

texuf:
...for request"

Add fix for "No Application Found Unable to find a matching application for request"

### DIFF
--- a/docs/web/fastcgi/nginx.md
+++ b/docs/web/fastcgi/nginx.md
@@ -41,6 +41,11 @@ Now the Nginx part is complete, but to finish the whole thing Mono FastCGI serve
 
      fastcgi-mono-server2 /applications=www.domain1.xyz:/:/var/www/www.domain1.xyz/ /socket=tcp:127.0.0.1:9000
 
+Note: On some systems this command doesn't work - you will get a "No Application Found
+Unable to find a matching application for request" Error  - removing the name (as demonstrated below) fixes the problem
+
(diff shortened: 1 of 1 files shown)


akoeplinger:
Hmm, I'm curious why this happens on "some systems", maybe a bug?

texuf:
I stumbled on the answer today - that param should be the name specified in the "location" field in the server config. In the example, the location is "/". If that location was changed to "/subdomain" then the resulting command would be .../applications=/subdomain:/var/www/www.domain1.xyz/... 

I think the documentation is just wrong.

akoeplinger:
Hmm, judging from the [FastCGI page](http://www.mono-project.com/docs/web/fastcgi/#how-applications-are-handled-and-how-to-configure-them) the existing line should actually work. `www.domain1.xyz:/:/var/www/www.domain1.xyz/` specifies `www.domain1.xyz` as the hostname, `/` as the virtual root/location and `/var/www/www.domain1.xyz` as the physical path.

The example you added should actually be equivalent to that, except that it listens on all hostnames (note there are only two colon-separated values, instead of three). So it might have something to do with how you access the resource, or rather what the `Host` HTTP-Header specifies. So if you just request http://localhost it won't work, as the fastcgi server only listens for www.domain.xyz.

Can you test whether this is true? If yes, we should probably just add your example as another way to listen on all hostnames.

akoeplinger:
I added some clarification and your example in  https://github.com/mono/website/commit/cb134c656279eef4f72893eeab5d44e99983b940.

PR #49, opened 2014-09-29
Embedded Mono: "mono-2.0.dll can not be located"

gumartinm:
Outdated documentation about how to compile and link Mono on Windows.

See: http://stackoverflow.com/questions/25958801/embed-mono-cannot-find-mono-2-0-dll
     http://stackoverflow.com/questions/22961704/getting-compiling-mono-2-0-dll-for-embedding


docs/advanced/embedding/index.md
```diff
@@ -92,7 +92,14 @@ On windows you need to generate an import library for mono.dll by getting the fo
 
      http://github.com/mono/mono/blob/master/msvc/mono.def
 
-and creating mono.lib with the command:
+Then in the bin directory of your mono installation folder (MONO_INSTALLATION_PATH\bin\) you should see some of these libraries: libmonoboehm-2.0.dll, libmonosgen-2.0.dll, libmono-2.0.dll or mono-2.0.dll. Depending on your mono's version you can see different names (in the current versions you will find [libmonoboehm-2.0.dll](/docs/advanced/garbage-collector/sgen/) and [libmonosgen-2.0.dll](/docs/advanced/garbage-collector/sgen/)) you must choose one of them and edit the mono.def file. For example, if you choose libmonosgen-2.0.dll the mono.def file should end up like this:
```
Therzok: Please remove the link to SGen from the Boehm mentioning.

akoeplinger:
I'm not familiar with Mono embedding, but I wonder if [monosgen.def](https://github.com/mono/mono/blob/master/msvc/monosgen.def) is actually the file that should be used for sgen instead of modifiying the mono.def file?

@vargaz can you clarify?

gumartinm:
There are differences between SGen and Boehm (at least for Mono 3.2.6) but they are related to the GC implementation (as expected):

```
me@myserver:~$ nm -D /path/to/library/libmonosgen-2.0.so | grep "mono_" | cut -d" " -f3 > outputsgen
me@myserver:~$ nm -D /path/to/library/libmonoboehm-2.0.so | grep "mono_" | cut -d" " -f3 > outputboehm
me@myserver:~$ diff outputsgen outputboehm
368,372d367
< mono_gc_register_bridge_callbacks
< mono_gc_scan_for_specific_ref
< mono_gc_toggleref_add
< mono_gc_toggleref_register_callback
< mono_gc_wait_for_bridge_processing
```

Besides, it seems as if [monosgen.def ](https://github.com/mono/mono/blob/master/msvc/monosgen.def) and [mono.def](https://github.com/mono/mono/blob/master/msvc/mono.def) are outdated. There are missing functions but I am not sure if they are important (it is mostly internal Mono stuff)

vargaz:
Probably monosgen.def should be used for sgen, and mono.def for boehm.

akoeplinger:
@madewokherd I see you added the monosgen.def file in mono/mono#837, can you confirm?

madewokherd:
I did not add monosgen.def, I just made it possible to update it in an automated way by running 'make update-def' in the msvc directory on a Linux machine with a build. If the library name in those .def files needs updated, the appropriate place for that is in the update-def target in msvc/Makefile.am.

Why is it necessary to generate the .lib file manually? It should be built automatically by MSVC with your .dll. If you compiled with GCC then there should be a way to get a .lib file out of that process.

Currently mono.def is for boehm and monosgen.def is for sgen (my PR did not change that), but I don't know any reason why that couldn't be changed.

akoeplinger:
@madewokherd sorry, didn't look beyond the PR ;-)

> Why is it necessary to generate the .lib file manually? It should be built automatically by MSVC with your .dll. If you compiled with GCC then there should be a way to get a .lib file out of that process.

This only applies when you self-built Mono, but when you're using the binary package I think you only have the DLL and need to generate the .lib file.

madewokherd:
Then I have to ask: who is building the binary packages, and why haven't they included the .lib files?

akoeplinger:
Seems like no one is currently responsible for the Windows packages, so in the meantime we should maybe just fix the Makefile you mentioned so it actually generates the correct library name in the .def files.

@gumartinm Can you check the .def file works "out of the box" when you make the following changes to [Makefile.am](https://github.com/mono/mono/blob/master/msvc/Makefile.am):

``` diff
update-def:
- perl create-windef.pl mono.def ../mono/mini/.libs/libmonoboehm-2.0.so mono-2.0.dll
- perl create-windef.pl monosgen.def ../mono/mini/.libs/libmonosgen-2.0.so monosgen-2.0.dll
+ perl create-windef.pl mono.def ../mono/mini/.libs/libmonoboehm-2.0.so libmonoboehm-2.0.dll
+ perl create-windef.pl monosgen.def ../mono/mini/.libs/libmonosgen-2.0.so libmonosgen-2.0.dll
```

If that works, we'll just need to update the docs page to mention you can choose between boehm and sgen and point to the respective .def file links. And PR the above change to mono of course ;)

madewokherd:
Of course, this requires someone with a Linux build to actually update the files.

It'd also be nice to have someone verify that this change doesn't break the MSVC build. I think the library name in the .def files is ignored when building with MSVC, so it should be fine?

gumartinm:
@akoeplinger The changes work but I do not have time to verify whether they break the MSVC build. This weekend I will give it a try.

Should I close this PR? I can open a new one once I find out whether everything is alright or the MSVC build is broken. In the worst case (not being able to find out if those changes break the MSVC build) just editing the .def files with the right library names should also work and for sure it doesn't break anything.

akoeplinger:
@gumartinm I don't mind, you can also just let this PR open until you've checked it and then add further changes here :smile:

akoeplinger:
@gumartinm any updates? If you can't test on MSVC I think it's not a big deal, we can make the PR and see what the team says.

gumartinm:
@akoeplinger Sorry, this month I had to stay away from C# and Mono. Just wait one more week and I will update the PR.

directhex:
I will be sure to include .lib files with Mono for Windows when I get closer to a working installer.

directhex:
> Then I have to ask: who is building the binary packages, and why haven't they included the .lib files?

Can you let me know whether https://drive.google.com/a/xamarin.com/file/d/0Bz6-k9ELOQf3YTE1RHV3Y0dNaFU/view works better than the current 3.2.3 release?

akoeplinger:
@gumartinm Could you try how this applies now with the Mono 3.12 Windows release? I verified it includes the .lib files, so generating them shouldn't be necessary anymore (though I'm not sure if they reference the correct library name?).

akoeplinger:
This should not be necessary anymore since the recent Windows packages include the .lib files. I replaced the outdated docs with https://github.com/mono/website/commit/6630532cbaa92bd62613fdf39087ad4d4e525707.

PR #53, opened 2014-10-27
Updated downloads page and linux-install docs

eropple:
Updated downloads page and linux-install docs to use the keyserver and to explicitly note that it's not Debian-only.

### DIFF
--- a/download/index.html
+++ b/download/index.html
@@ -39,13 +39,13 @@ redirect_from:
         <div>
           <p>These packages are built, tested and distributed by Xamarin. Use these if you want to use a stable, official version of Mono.</p>
           <h5>Mono packages for Debian, Ubuntu, and derivatives</h5>
-          <p>The Mono package repository is as follows. Please see the installation guide for details on adding it:</p>
+          <p>While the APT package is built against Debian Wheezy, it is compatible with a number of Debian derivatives, including Ubuntu 14.04 LTS. See the <a href="http://www.mono-project.com/docs/getting-started/install/linux/">installation guide</a> for detailed installation instructions across many versions of Debian and Ubuntu.</p>
           <p>Apt repo: <tt>deb http://download.mono-project.com/repo/debian wheezy main</tt></p>
           <h5>Mono packages for Red Hat, SUSE, and derivatives</h5>
           <p>Yum repo: <tt>http://download.mono-project.com/repo/centos/</tt></p>
         </div>
         <h4>Linux distribution packages</h4>
-        <p>Mono is packaged in various Linux distributions by third-party maintainers. Some packages do not ship the latest Mono version, this depends on the distribution/maintainer.</p>
+        <p>Mono is packaged in various Linux distributions by third-party maintainers. Some packages do not ship the latest Mono version; this depends on the distribution/maintainer.</p>
         <div>
           <ul>
             <li><a href="https://www.archlinux.org/packages/extra/i686/mono/">Arch Linux</a></li>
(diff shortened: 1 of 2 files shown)


akoeplinger:
@directhex what do you think about this?

download/index.html
```diff
@@ -39,13 +39,13 @@
         <div>
           <p>These packages are built, tested and distributed by Xamarin. Use these if you want to use a stable, official version of Mono.</p>
           <h5>Mono packages for Debian, Ubuntu, and derivatives</h5>
-          <p>The Mono package repository is as follows. Please see the installation guide for details on adding it:</p>
+          <p>While the APT package is built against Debian Wheezy, it is compatible with a number of Debian derivatives, including Ubuntu 14.04 LTS. See the <a href="http://www.mono-project.com/docs/getting-started/install/linux/">installation guide</a> for detailed installation instructions across many versions of Debian and Ubuntu.</p>
```
akoeplinger: The heading already says "Debian, Ubuntu, and derivatives" so I'm not sure we need to duplicate this here.

Also, remove the "http://www.mono-project.com" prefix from the link.

download/index.html
```diff
@@ -39,13 +39,13 @@
...
           <p>Apt repo: <tt>deb http://download.mono-project.com/repo/debian wheezy main</tt></p>
           <h5>Mono packages for Red Hat, SUSE, and derivatives</h5>
           <p>Yum repo: <tt>http://download.mono-project.com/repo/centos/</tt></p>
         </div>
         <h4>Linux distribution packages</h4>
-        <p>Mono is packaged in various Linux distributions by third-party maintainers. Some packages do not ship the latest Mono version, this depends on the distribution/maintainer.</p>
+        <p>Mono is packaged in various Linux distributions by third-party maintainers. Some packages do not ship the latest Mono version; this depends on the distribution/maintainer.</p>
```
akoeplinger, line 48: Any specific reason for this change? Is there anything wrong with using a comma? (I'm no native speaker, so please explain :smile: )

download/index.html
eropple: As discussed with @directhex on Twitter just now, I would argue that it's not clear - I'm an experienced Linux user and went "OK, that's the Wheezy link", went ahead and dropped the appropriate ubuntu/trusty bits into the line, and it blew up.

When you're not acting as a user would expect, it strikes me that you should make that very explicit.

download/index.html
eropple, line 48: It's a run-on sentence - two independent clauses bolted together by a comma. Commas are used for relating dependent clauses and an independent clause.

http://en.wikipedia.org/wiki/Comma_splice

download/index.html
akoeplinger: Good point. Though without the key it'll not work anyway, so maybe it'd be better to just nuke this section and only link to the install guide where everything is explained in more detail?

download/index.html
akoeplinger, line 48: Thanks, learning never stops ;-)

download/index.html
eropple: It works (on Ubuntu) without the key, it just whines at you about being unverified.

I think there's an argument for moving the interesting parts of the install guide onto this page to help get people going more quickly (maybe a Debian/Ubuntu Latest/Ubuntu Old tab bar?) but I don't know enough about your UX guidelines to say. :)

akoeplinger: Hm, I think if people have to click one more time to get to the content I'm fine with that rather than exposing semi-complete info on this page.

Here's what I have in mind, do you think that'd be a good compromise (the links point to the respective sections in the install guide)?
screen

akoeplinger:
@eropple I've pushed a slightly modified version. Please check it out and let me know if there are any problems with it :)

PR #60, opened 2014-11-15
Correct default xsp4 port number

DasAllFolks:
The tutorial implies that the sample .aspx page should run on port 9000; however, it appears that xsp4 currently uses port 8080 for this purpose.

To quote from xsp4 --help:

$ xsp4 --help
XSP server is a sample server that hosts the ASP.NET runtime in a
minimalistic HTTP server

Usage is:

```
xsp.exe [...]

--port N: n is the tcp port to listen on.
                Default value: 8080
```

[...]

### DIFF
--- a/docs/getting-started/mono-basics.md
+++ b/docs/getting-started/mono-basics.md
@@ -89,7 +89,7 @@ Then run the xsp4 command from that directory:
 xsp4
 ```
 
-Use a web browser to contact [http://localhost:9000/hello.aspx](http://localhost:9000/hello.aspx)
+Use a web browser to contact [http://localhost:8080/hello.aspx](http://localhost:8080/hello.aspx)
 
 Gtk# Hello World
 -----------------


akoeplinger:
Hmm, it runs on port 9000 by default for me:

```
alexander@ubuntu-desktop:~/Desktop/test$ xsp4
xsp4
Listening on address: 0.0.0.0
Root directory: /home/alexander/Desktop/test
Listening on port: 9000 (non-secure)
Hit Return to stop the server.
```

Maybe you have an older xsp version?

akoeplinger:
I just confirmed this is from an old xsp, e.g. back in 2013 it used 8080: [main.cs#L101](https://github.com/mono/xsp/blob/b7ba3073fc99dc62f77898b1fb57fda324c1dcfc/src/Mono.WebServer.XSP/main.cs#L101), but current master uses 9000: [ConfigurationManager.cs#L47](https://github.com/mono/xsp/blob/8a31bc625727594d42f94173768bee5cf8afd0a4/src/Mono.WebServer.XSP/ConfigurationManager.cs#L47).

I'm closing this as it doesn't apply to current xsp, but thanks for taking the time to bring this up :smile:

DasAllFolks:
Hey Alexander,

Thanks for getting back about this so quickly!

I actually experienced this issue with the [very latest Mono MDK .pkg available for Mac OS X](http://www.mono-project.com/download/) (I'm running Mavericks, if that gives you better context information):  I only downloaded and installed this MDK yesterday, but it was (and still is) definitely running on port 8080 by default!

Perhaps that change to the default xsp port number hasn't yet been ported to xsp4 in the Mac MDK .pkg installers?...

akoeplinger:
Ah, yeah sorry, the Mac package is a different beast, looks like they ship an old xsp version. 

Actually, they seem to have upgraded xsp for Mac a few days ago (https://github.com/mono/bockbuild/commit/0f6dcf8d5e989562a92bbc055837224fecad7f79), so it should ship with the next Mono release (or the one after that, I believe 3.12 is already freezed).

DasAllFolks:
Great, thanks Alexander!  Already looking forward to trying out 3.12 once it ships. :-)

akoeplinger:
Might be a good idea to just alter the command on the page to `xsp4 --port=9000`, so it'll work on every version. Can you do that?

DasAllFolks:
Sure!  Doing this now.

DasAllFolks:
Okay, [fresh pull request is up for review](https://github.com/mono/website/pull/61).

Closing this pull request for now, as it's now stale.

Thanks, Alex!

PR #61, opened 2014-11-16
Explicitly provided --port flag in xsp4 walkthrough

DasAllFolks:
This should (hopefully) prevent some novice headaches, given that the default xsp4 port number apparently varies from download to download right now (e.g., Mac OS X MDK .pkg 3.10 still uses port 8080 as default).


akoeplinger:
Great, thanks!

PR #62, opened 2014-11-17
Improve LLDB script.

Therzok:


akoeplinger:
Do you want this to be reviewed by someone or is it good to merge?

Therzok:
I don't think it needs reviewing.

I mapped c++ API to python API where applicable.

Added 'all' and '#frames' switches, similar to 'bt'.

Added \* in front of current thread and frame.

Only problems are:
Thread header isn't the same as BT. (Doable, but meh)
Extra newline at the end of 'monobt all'. Not really annoying.

PR #65, opened 2014-11-19
Removing wxNET as it's Dead

thiagojedi:
The wxNET description was duplicated. One here and another one in Dead Toolkits page. As the Project is still dead (no updates since 2011), it shouldn't be here.


akoeplinger:
Good catch, thanks!

PR #69, opened 2014-11-27
Add sudo to the shell commands

obadz:
..so that they can be readily pasted

### DIFF
--- a/docs/getting-started/install/linux/index.md
+++ b/docs/getting-started/install/linux/index.md
@@ -14,16 +14,16 @@ Regardless of your distribution, you will need the Mono Project GPG signing key,
 Debian, Ubuntu, and derivatives
 -------------------------------
 
-Add the Mono Project GPG signing key **in a root shell** with:
+Add the Mono Project GPG signing key with (if you don't use sudo, be sure to switch to root):
 
 ``` bash
-apt-key adv --keyserver pgp.mit.edu --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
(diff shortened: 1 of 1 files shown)


txdv:
Depends on the system. Easier to use doesn't mean safer to use.

akoeplinger:
I'm a bit torn on this one. Debian doesn't have sudo by default (so users there can't copy/paste either), which means the only real argument in favor of sudo I have is that Ubuntu is the go-to distribution for users that are new to Linux, so they might not immediately know what "run this in a root shell" means.

Debian users on the other hand will more likely know that they need to remove `sudo` from commands and switch to a root shell instead.

I guess I'm slightly :+1: for taking this change, what do others think?

obadz:
I think sudo has become the standard way to encode into publicly posted commands that they should be run as root. Ubuntu & Mint ship it by default and I think Debian has it in the standard system utilities task? (isn't that installed unless specifically unchecked by user during install?)

docs/getting-started/install/linux/index.md
```diff
@@ -14,16 +14,16 @@ Regardless of your distribution, you will need the Mono Project GPG signing key,
 Debian, Ubuntu, and derivatives
 -------------------------------
 
-Add the Mono Project GPG signing key **in a root shell** with:
+Add the Mono Project GPG signing key with (if you don't use sudo, be sure to switch to root):
```
akoeplinger: any reason for `with`? I think it reads better without.

akoeplinger:
Please also change the other occurences of > to | sudo tee.

akoeplinger:
Thanks! If you have time, you could make the same changes to the MonoDevelop guide, to keep everything consistent: https://github.com/mono/md-website/blob/gh-pages/download/linux.md

PR #70, opened 2014-12-02
Update AssemblyFactory references to use AssemblyDefinition

gpduck:
`AssemblyFactory` has been removed, so I updated the FAQ to reference the current method for reading/writing assemblies.


akoeplinger:
Thanks!

PR #73, opened 2014-12-03
Fixed building on OSX quick script

jonorossi:
At least on Yosemite I needed to build m4 before I could build autoconf, and to build automake I needed autoconf on the path.

I'm just starting with building mono, so reject these changes if they are wrong.

### DIFF
--- a/docs/compiling-mono/linux/index.md
+++ b/docs/compiling-mono/linux/index.md
@@ -65,12 +65,11 @@ If you are on a Debian system (Debian/Ubuntu) and you just want to install Mono 
 #!/bin/bash
 
 PREFIX=$@
-
 if [ -z $PREFIX ]; then
-  PREFIX="/opt/mono/"
+  PREFIX="/usr/local"
 fi
 
-# Ensure you have write permissions to /opt/mono/
+# Ensure you have write permissions to PREFIX
 sudo mkdir $PREFIX
 sudo chown -R `whoami` $PREFIX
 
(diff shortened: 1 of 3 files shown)


akoeplinger:
Assuming the changes are correct, could you make them to the "Building Mono Dependencies" section too?
I currently don't have a Mac for testing, but I'm going to simply trust you on this :smile:

jonorossi:
@akoeplinger sorry for missing that section, I've updated the pull request.

I just realised why it worked for the author, the examples use `PREFIX=/usr/local` which the third paragraph of the page strongly advises against... i.e. it is already on the path. I used `PREFIX=/mono`, would you like me to change that too? The Linux page says to use something like `/mono` too, but then goes and uses `/opt/mono`. There is a lot of different advice being thrown around here, but `/opt/mono` sounds like a good place for it. Let me know if you want me to fix it.

akoeplinger:
@jonorossi `/usr/local` is different from `/usr` (the latter is advised against, as that is where the Mono packages are installed). The benefit of having it in `/usr/local` as you said is that it is already in the PATH, so you don't need to adjust it afterwards to use the custom-built Mono (e.g. like detailed in [Parallel Mono Environments](http://www.mono-project.com/docs/compiling-mono/parallel-mono-environments/)

I agree that the script should work for other prefixes though, so if your changes allow this then fine.

I think we should standardize on the following:
1. default prefix is `/usr/local`
2.  for other prefixes like `/opt/mono` we should point to the Parallel Mono Environment document

jonorossi:
@akoeplinger thanks, I understand now why only `/usr` local is advised against, I just incorrectly assumed it also meant `/usr/local` because they are both on the path.

I've updated the linux script to standardise on `/usr/local` and both OSX and Linux pages already link to the parallel mono environments page, however I've got two quick questions on best practice:
1. Is it a good idea to run `sudo chown -R`whoami`/usr/local`, could that break anything?
2. Is it a good idea for the custom built mono to always live on the path with the OS installed mono? i.e. are we increasing the changes something will break with two `mono`s on the OS path?

akoeplinger:
@jonorossi 2. should be harmless since`/usr/local` should come before `/usr` in PATH.
For 1. I have no idea, I didn't write that script.

jonorossi:
@akoeplinger many thanks for helping me "get this" stuff.

From what I've read it shouldn't do any harm changing the owner of `/usr/local` from root to the current user and since the page already had that I think these changes are ready to merge if you are still happy with them.

akoeplinger:
Great, thanks!

PR #74, opened 2014-12-08
Spelling error fix

eje211:
"Principal" and "principle" mean different things. The wrong one was used in this text.


akoeplinger:
Good catch, thanks!

PR #82, opened 2015-01-04
Update nginx.md

techarch:
The current version of fastcgi-mono-server is 4 so line 42 needs to reflect that with: fastcgi-mono-server4


akoeplinger:
Thanks!

PR #83, opened 2015-01-05
Finished index.md for the advanced section of the doc 

peteryongzhong:


akoeplinger:
Thanks! Can you please change the URLs to start with "/docs/advanced" (i.e. being relative to the root instead of the current document). This is the style we use everywhere else.

peteryongzhong:
my apologies, I will change it now

peteryongzhong:
there, fixed

PR #84, opened 2015-01-09
Update index.md

aman-saggu-git:


akoeplinger:
I'm going to merge this, but we'll need to do some cleanup on that page anyway since it's quite outdated.

PR #86, opened 2015-01-10
Changes to docs/index.md and news/index.html

aman-saggu-git:
In news/index.html there was need of line break before first news and I put a horizontal line at end of each News

In docs/index.md I have  placed the contents inside accordion in table

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -19,47 +19,73 @@ Our website is open source on [GitHub](https://github.com/mono/website). If you 
     <dd class="accordion-navigation">
         <a class="panel" href="#getting-started"><h2>Getting Started</h2></a>
         <div id="getting-started" class="content active">
-            <ul class="disc">
-                <li><a href="/docs/getting-started">Overview</a></li>
-                <li>Install Mono on <a href="/docs/getting-started/install/mac/">Mac OS X</a>,<a href="/docs/getting-started/install/linux/">Linux</a>,<a href="/docs/getting-started/install/windows/">Windows</a></li>
-                <li><a href="/docs/getting-started/mono-basics/">Mono Basics</a></li>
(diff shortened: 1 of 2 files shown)


akoeplinger:
> In news/index.html there was need of line break before first news and I put a horizontal line at end of each News

The line break you put in would be repeated for every item, not just the first, you need to move it outside the foreach. Also, the `<hr/>` should be after the closing `</article>` tag.

> In docs/index.md I have placed the contents inside accordion in table 

I really dislike tables. Can you explain why this is necessary?

aman-saggu-git:
that was just for adding some simple and nice look to contents inside Accordion.

aman-saggu-git:
It looks like this
screenshot

aman-saggu-git:
if you don't like tables i can put panel class there what you suggest

aman-saggu-git:
After Panel it will look like this -

screenshot_panel

akoeplinger:
Yeah, the one with panel looks much better.

Can you separate the changes to news and docs? The two aren't related and should be in their own pull requests.

aman-saggu-git:
ok sure , i will do that with panels.
can you suggest me how can i put them in different pull request . not used git so much

akoeplinger:
It's a bit difficult now because you didn't use a separate branch and did everything in gh-pages.
You'd need to reset your gh-pages branch to the original state (so it doesn't contain any of your commits), then create a new branch for the news changes, apply the changes, push the branch to GitHub and open a PR. Then do the same for the docs changes.

You may want to take a look into using a GUI like [SmartGit](http://www.syntevo.com/smartgit/) or [SourceTree](http://www.sourcetreeapp.com/) if you're having trouble understanding the git command line and reading up some tutorials about branching in git.

aman-saggu-git:
ok, thanks I am looking at these things

PR #98, opened 2015-02-19
Remove extraneous backticks in Mono.Cairo tutorial

adamburgess:
Also fixes a small syntax error: Format.RG**V**24 -> Format.RG**B**24


akoeplinger:
Thanks!

PR #103, opened 2015-03-08
Add Homebrew package names

Daniel15:
Adds instructions for installing dependencies via Homebrew.

### DIFF
--- a/docs/compiling-mono/mac/index.md
+++ b/docs/compiling-mono/mac/index.md
@@ -43,6 +43,12 @@ for i in *.tar.gz; do tar xzvf $i; done
 for i in */configure; do (cd `dirname $i`; ./configure --prefix=$PREFIX && make && make install); done
 ```
 
+Alternatively, dependencies can be installed through Homebrew:
+
+``` bash
+brew install autoconf automake libtool pkg-config glib libmcrypt freetype cmake libmemcached
+```
+
 Building Mono from a Release Package
 ------------------------------------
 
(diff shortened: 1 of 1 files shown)


docs/compiling-mono/mac/index.md
```diff
@@ -43,6 +43,12 @@ for i in *.tar.gz; do tar xzvf $i; done
 for i in */configure; do (cd `dirname $i`; ./configure --prefix=$PREFIX && make && make install); done
 ```
 
+Alternatively, dependencies can be installed through Homebrew:
+
+``` bash
+brew install autoconf automake libtool pkg-config glib libmcrypt freetype cmake libmemcached
```
akoeplinger: are all of those dependencies really necessary? e.g. what are cmake, libmemcached used for?

Therzok: Actually, you can get the dependencies list with `brew deps mono`. I think it would be something like `brew install $(brew deps mono)`

Daniel15: @akoeplinger - I kept instaling dependencies until the build succeeded. Without libmemcached I got an error saying it couldn't find `libmemcached/aclocal.m4`. However I just tried it again and it worked this time even after uninstalling libmemcached so I'm not sure why it complained initially.

@Therzok `brew deps mono` gives me an empty result.

akoeplinger:
I tested on a clean Yosemite and was able to compile with just `brew install autoconf automake libtool pkg-config` so I squashed and merged this in with: https://github.com/mono/website/commit/48a33d11a078188d7838e0ee7a029b7e8d15f59f. Thanks!

PR #124, opened 2015-06-11
Additional package needed

Dr4ik:
Without libtool-bin autogen will complain (tested on Ubuntu 15.04):
**Error**: You must have `libtool' installed to compile Mono.

### DIFF
--- a/docs/compiling-mono/linux/index.md
+++ b/docs/compiling-mono/linux/index.md
@@ -22,7 +22,7 @@ On Debian based distribution you should guarantee that some packages are install
 This can be done easily by using *apt-get*:
 
 ```bash
-  $ sudo apt-get install git autoconf libtool automake build-essential mono-devel gettext
+  $ sudo apt-get install git autoconf libtool libtool-bin automake build-essential mono-devel gettext
 ```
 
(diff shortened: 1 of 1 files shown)


akoeplinger:
Hmm, I'd have thought this would be pulled in by `libtool`?

Dr4ik:
I've just tried to build it on Ubuntu 15.04.
Maybe I'm the only one who had this problem :)

akoeplinger:
Interesting. I'll take a look ;)

akoeplinger:
So, the issue is that libtool was split: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=761791
Ideally, we should move to libtoolize, but until that happens, we need `libtool-bin`.

The problem is that installing `libtool-bin` doesn't work on earlier Ubuntu releases. Can you modify your change into a "Note: on Ubuntu 15.04, install libtool-bin ..." instead?

Dr4ik:
Done.
I hope i've done it right :)

Dr4ik:
I hope now it looks OK.

akoeplinger:
Thanks!

PR #133, opened 2015-07-17
Update index.md

mocsy:
Mono.Data.SybaseClient is not part of mono anymore, this documentation is somewhat misleading.
mono/old-code@9db16a8


akoeplinger:
Thanks! Guess we should move the sybase page itself to /archived then as well.

PR #141, opened 2015-09-01
Add info about bug: 21810.

mfilippov:


akoeplinger:
As discussed in the bugzilla ticket, this was fixed in 4.2, merging. Thanks!

PR #144, opened 2015-09-08
As of Mono 4.0, C# 6.0 is supported

croconoster:
http://www.mono-project.com/docs/about-mono/releases/4.0.0/ release notes specifies that C# 6.0 has been integrated into Mono. Not sure if we need to remove the C# 5.0 line in compatibility.


akoeplinger:
I think it'd make more sense in a (new) .NET 4.6 section, as that's the release with which C#6 got shipped by MS.

croconoster:
Removed C# 6.0 from .NET 4.5 and added it to a new section called .Net 4.6

akoeplinger:
Thanks!

PR #145, opened 2015-09-08
From Mono 4.2 will be supported

croconoster:
There are a bunch of fixes in Mono 4.2 which deals with MVC 5 support. Currently have MVC 5 running on mono with exception to the fixes I'm waiting for. The only thing that was not working was the lowercase URL's and the append tailing slash function was missing. This should now be fixed.

### DIFF
--- a/docs/about-mono/compatibility.md
+++ b/docs/about-mono/compatibility.md
@@ -19,6 +19,7 @@ Here is a slightly more detailed view, by .NET framework version:
 |<i class="fa fa-check"/>|C# 5.0 - async support|
 |<i class="fa fa-check"/>|Async Base Class Library Upgrade|
 |<i class="fa fa-exclamation-triangle"/>|MVC4 *- Partial, no async features supported.*|
+|<i class="fa fa-check"/>|MVC5 |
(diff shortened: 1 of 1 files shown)


akoeplinger:
> Currently have MVC 5 running on mono with exception to the fixes I'm waiting for.

So does that mean MVC5 works on Mono 4.2 or not? It's currently in the alpha channel, so if there are fixes that aren't included in the alpha yet they'd need to be backported.

croconoster:
MVC 5 works on Mono 4.0.3 apart from the appendtailingslash attribute which is missing and lowercase url's. This has been fixed in 4.2 which sits in the alpha channel. https://github.com/mono/mono/commit/58e0608c10c621b15fd9f396f8e46c567891b231 is where MVC5 missing methods were fixed.

Probably a good idea to hold this pull request until 4.2 makes it into Production release.

akoeplinger:
OK. I think the same async restrictions from the MVC4 line above apply though, can you add those here as well?.

croconoster:
Updated. Think you are correct.

PR #147, opened 2015-09-13
Added error message for libtool

martinjt:
This should hopefully mean it get's picked up by the search engines

### DIFF
--- a/docs/compiling-mono/linux/index.md
+++ b/docs/compiling-mono/linux/index.md
@@ -25,7 +25,9 @@ This can be done easily by using *apt-get*:
   $ sudo apt-get install git autoconf libtool automake build-essential mono-devel gettext
 ```
 
-Note: if you are using Ubuntu 15.04 or later, you also need to install the `libtool-bin` package.
+Note: if you are using Ubuntu 15.04 or later, you also need to install the `libtool-bin` package.  Without it, you will get the following error:
+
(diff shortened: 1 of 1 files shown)


docs/compiling-mono/linux/index.md
```diff
@@ -76,7 +78,7 @@ sudo mkdir $PREFIX
 sudo chown -R `whoami` $PREFIX
 
 # Ensure that all required packages are installed.
-sudo apt-get install git autoconf libtool automake build-essential mono-devel gettext
+sudo apt-get install git autoconf libtool libtool-bin automake build-essential mono-devel gettext
```
akoeplinger, line 81: As noted in https://github.com/mono/website/pull/124#issuecomment-111735382, this won't work on earlier Ubuntu releases. Can you please remove it from here?

martinjt, line 81: What about adding an if statement based on the version?

akoeplinger, line 81: Sure. Just make sure that `lsb_release` is actually available on the system before trying to use it, as it might not be available on some Debian versions out of the box (or ignore any errors from it).

akoeplinger:
I merged the first part with https://github.com/mono/website/commit/98e729b61d7060085ad052f4b4e105dffa0d26bc. Feel free to add additional improvements as discussed in follow up PRs. Thanks!

PR #153, opened 2015-10-17
Williamsrz patch /supported platforms osx

wsrzx:
tradução da página


akoeplinger:
Did you mean to submit this to https://github.com/MonoBrasil/website instead of here? If not, please resubmit with only the relevant changes :)

PR #156, opened 2015-11-03
Update index.md

Baltasarq:
Fixed various problems:
- Cairo.Graphics is obsolete, you have to use Cairo.Context instead.
- In order to get the Surface (the target of the Context), you call GetTarget(), Target (property) is obsolete.
- You don't need to cast to IDisposable in order to call Dispose(). In any case, the name of the reference pointing to the Cairo.Context should match with the one calling to Dispose(). I have renamed all of them to just "g", instead of "context", "gr" and "g".
- The hint in which Gdk's drawing functions are recommended for simple tasks is wrong, since these functions have disappeared in Gtk3.


akoeplinger:
Thanks!

PR #166, opened 2015-12-06
configure `--target=` vs `--host=`

sushihangover:
Correct 32-bit compile as using `--target=` instead of `--host=` will result in target name prefixes added to the llvm binaries, as if you were building a cross-compiled toolchain and thus mono's `configure` process to fail to find `llvm-config`.

```
checking for llvm-config... no
configure: error: llvm-config not found.
```


akoeplinger:
Looks like we're using `--build` in our packaging, so I'd rather switch the docs to that: https://github.com/mono/bockbuild/blob/master/packages/mono-llvm.py#L21

PR #171, opened 2016-01-18
Fix some links

tezaurismosis:
Fix link to PPC topic from IBM
Fix link to elfspec_ppc.pdf


akoeplinger:
Hmm, not sure linking to some random sourceforge repo is better in this case. Maybe https://developer.apple.com/library/mac/documentation/DeveloperTools/Conceptual/MachORuntime/ ?

Thanks for the corrections!

tezaurismosis:
I can't find pdf-file on official sources, but topic not equal to original specification (92 pages long). So, we can add link to topic at least as a start point to exploration.

akoeplinger:
Sure, feel free to add that in another PR :)

tezaurismosis:
Ok :)

PR #177, opened 2016-02-06
Autofocus right OS with JavaScript

txdv:
It automatically focuses the right tab according to the windows.navigator variable!

### DIFF
--- a/download/index.html
+++ b/download/index.html
@@ -105,3 +105,23 @@ redirect_from:
     <p>To try the latest planned changes, check the <a href="/docs/getting-started/install/weekly-packages/">weekly packages</a> download page.</p>
   </div>
 </div>
+
+<script>
+function runScripts() {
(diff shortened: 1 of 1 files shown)


txdv:
http://stackoverflow.com/questions/7044944/jquery-javascript-to-detect-os-without-a-plugin

akoeplinger:
Looks nice, thank you!

One question: can we make it so that it doesn't switch if I'm linking to the tab directly like http://www.mono-project.com/download/#download-win even if I'm on another OS?

txdv:
Yeah sure, obvious mistake, was sleepy yesterday so I just pushed what I had and went to sleep.

akoeplinger:
Merged, thank you! I made one small correction which I pushed in https://github.com/mono/website/commit/c35a7c5b54a33f78096589f1f4315fa939adf62c :)

PR #178, opened 2016-02-09
Update linux compiling-mono instructions

timjk:
- No apparent location for packages on the 'Mono website'
- No configure script on the downloaded package (only configure.ac)
- Other code snippets use ./autogen --prefix=$PREFIX

### DIFF
--- a/docs/compiling-mono/linux/index.md
+++ b/docs/compiling-mono/linux/index.md
@@ -30,14 +30,14 @@ Note: if you are using Ubuntu 15.04 or later, you also need to install the `libt
 Building Mono from a Release Package
 ------------------------------------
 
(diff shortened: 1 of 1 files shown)


akoeplinger:
That section is talking about the Tarballs, which contain a `configure` script (i.e the output of the autogen.sh and configure.ac scripts) and don't rely on autotools being installed:

image

The zip and .tar.gz files you can download from GitHubs UI are not complete (they miss submodules) and aren't usable for this case.

akoeplinger:
I do think we can make the current description clearer by mentioning "tarball" etc. explicitly to avoid confusion. I'm going to close this PR as it isn't applicable now, but please feel free to discuss further or send a new PR :) Thank you!

timjk:
Ahhhhhhhh, thanks @akoeplinger.

PR #181, opened 2016-02-10
Update 4.2.1 release notes re: PasswordExtendBytes incompatibility

xmcclure:
Attempting to address bugzilla #38451.

Note this includes a small downloadable zip containing a workaround for the PasswordExtendBytes compatibility issue. This means there's now a zip sitting in the releases directory. If there's a better way to embed that file let me know.


xmcclure:
Thumbs up from Sebastien Pouliot. Hopefully the travis failure is spurious…?

PR #188, opened 2016-03-12
Updated Fedora version

yotamN:
It's a small fix but I think it is important to keep it updated, when I first got to this page I thought it was old so I installed from the distribution package instead. I don't know if Fedora 19 is still supported but Fedora 22 is the latest version that still get updates so there is no reason someone will use something older.


akoeplinger:
Thanks!

PR #204, opened 2016-05-06
Add note about adding certs with mozroots tool

cklanac:


akoeplinger:
Hmm, running mozroots shouldn't be necessary any more after the cert-sync changes in Mono 3.12: http://www.mono-project.com/docs/about-mono/releases/3.12.0/#cert-sync

In which context did you run into this issue?

PR #206, opened 2016-05-16
Update coop-suspend.md

luhenry:


docs/advanced/runtime/docs/coop-suspend.md
```diff
@@ -145,16 +145,18 @@ Creates a C lexical scope. It causes a transition from Unsafe to Safe mode.
...
 ```c
 MonoArray *x;
 int res;
 MONO_PREPARE_BLOCKING
-res = read (1, mono_array_addr (x, char, 0), mono_array_length (x), 0);
+res = read (1, mono_array_addr (x, char, 0), mono_array_length (x), 0); // if a GC run while read is blocked in the OS, the object x might be moved, and x would then point to garbage, or worst, in the middle of another object. And whenever the OS would write into the buffer passed to read, it would override managed memory.
 MONO_FINISH_BLOCKING
 ```
 
+To use an object reference in a blocking section, you need to pin the object in the managed heap with a GC handle.
```
kumpera: That's not 100% accurate.
We allow objects to leak into GC_SAFE regions as long as they are externally pinned and code doesn't touch ref fields.

akoeplinger:
should we merge this?

PR #207, opened 2016-05-16
Add a Coop state transition diagram

lambdageek:
Perhaps this will help to explain what's going on within a thread.


docs/advanced/runtime/docs/coop-state-machine.dot
```diff
@@ -0,0 +1,114 @@
...
+	{
+		edge [label = "detach"];
+		{ Running Blocking } -> Detached ;
+		// Async_Suspend_Requested -> Async_Suspend_Requested;
+	}
+	
+
+	// do blocking
+	{
+		edge [label = "do_B", color = "green", fontcolor="darkgreen"];
```
luhenry, line 53: why `do_B` and not `do_blocking`?

lambdageek, line 53: Because I couldn't figure out how to get Graphviz not to overlap labels if they get too wide. 😢

kumpera:
The self requested and async requested are inverted. We can't upgrade a self request to async request, only the other way around.

docs/advanced/runtime/docs/coop-suspend.md
```diff
@@ -68,6 +68,8 @@ states:
 - Blocking: The current thread is executing code that won't touch managed memory, so it can be assumed to be suspended
 - Blocking and Suspended: The current thread finished executing blocking code but there was a pending suspend against it, it's waiting to be resumed
 
+![Coop state machine transition diagram](coop-state-machine.png)
```
akoeplinger: You need to put the link as `/docs/advanced/runtime/docs/coop-state-machine.png`, because the coop-suspend.md gets turned into coop-suspend/index.html so the relative link would search in that directory and not find the image there.

PR #209, opened 2016-05-25
Remove broken image reference of: buttons.png

jasonsturges:
Proposing removal of broken image reference `[Image:buttons.png]`, which doesn't appear to be GitHub flavored markdown.

This image asset cannot be found within this repository; nor, is present within the [earliest web archive of this page](https://web.archive.org/web/20051230200243/http://www.mono-project.com/GtkSharp:_Buttons).  Only artifacts that can be located are perhaps from the former CVS repository.


akoeplinger:
Thank you!

PR #211, opened 2016-06-06
Fix/document download location messiness

oskarb:
Update references to out-of-date versions. Also, not all version are available at GNOME FTP - at least document and link to some additional sources.


akoeplinger:
Thank you!

PR #217, opened 2016-06-21
use jekyll-compress-html v3.0.2 to minify html output

PeterDaveHello:


akoeplinger:
How much does this save or improve the page speed? I'm hesitant adding this kind of "magic" to the site as it'll make certain kinds of issues harder to track down.

PeterDaveHello:
Saved about 20%, for developer working on development works can temporarily remove that layout for easier debug, users can benefit from the size decrease. This is no "magic" anyway, a common method to make the src smaller. Almost all the famous libraries also provide the minified js and css files.

Origin:

```
index: 8268
community: 7975
doc: 18267
```

Minified:

```
index: 6954
community: 6711
doc: 14737
```

akoeplinger:
GitHub Pages serves the HTML gzipped already, so I compared the byte counts without and with this patch:

Before:

```
index: 8340, 2750 (gzipped)
community: 8047, 2828 (gzipped)
doc: 18339, 4642 (gzipped)
```

After:

```
index: 7026, 2566 (gzipped)
community: 6783, 2659 (gzipped)
doc: 14809, 4373 (gzipped)
```

So there is nearly no impact on the gzipped size.

> This is no "magic" anyway, a common method to make the src smaller.

What I mean with "magic" are issues such as https://github.com/penibelst/jekyll-compress-html/issues/71 or https://github.com/penibelst/jekyll-compress-html/issues/70. Those are hard to notice and require ugly workarounds.

> Almost all the famous libraries also provide the minified js and css files.

The minified CSS/JS libraries we use are used by millions others too so we can rely on them being fine.

All in all I don't think this is worth the complexity given the little gain in effective page size reduction, hence I'm closing the PR. Thanks for the contribution!

PR #221, opened 2016-07-17
Add atomics and memory model doc

alexrp:


alexrp:
cc @kumpera @vargaz

docs/advanced/runtime/docs/atomics-memory-model.md
```diff
@@ -0,0 +1,128 @@
...
+These rules apply to high-level languages like C# and F# as they target the CLI.
+
+### `volatile.` prefix opcode in CIL
+
+When the `volatile.` prefix opcode is used in CIL, it imposes acquire/release semantics on the next non-prefix opcode.
+For loads, it results in acquire semantics. For stores, it results in release semantics.
+
+This prefix opcode has no effect on atomicity beyond the standard rules of the CLI.
+
+### `volatile` keyword in C#
```
akoeplinger: you need to write `C\#` so the # isn't treated as part of the markdown header

docs/advanced/runtime/docs/atomics-memory-model.md
```diff
@@ -0,0 +1,128 @@
...
+### `Volatile` class
+
+The methods on the `Volatile` class are all atomic regardless of system bitness, and result in acquire and release
+barriers for loads and stores respectively.
+
+The 64-bit methods on this class are not atomic with respect to loads or stores made through other means than the
+methods on this class and the `Interlocked` class. This is because such 64-bit operations may need to be implemented
+with a lock on 32-bit systems.
+
+The MSDN documentation lies about the C# compiler emitting calls to this class's methods. It does not.
```
akoeplinger: Can we file a bug e.g. on https://github.com/dotnet/core-docs to get this fixed?

alexrp: Are .NET BCL docs actually stored there? If so, I don't see it.

alexrp: BTW, it may actually be intentional in the sense that the (32-bit) `Volatile` methods are what the `volatile` keyword in C# is actually specified to match semantically. It's technically a bug that `Thread.VolatileRead`/`Thread.VolatileWrite` have stronger barriers than specified, and I don't think they want new code to rely on this.

akoeplinger: I don't think so, but at least the docs people monitor the repository and can redirect the bug to where it needs to be (e.g. I wouldn't know where to start filing an "internal" bug on MSDN 😄 )

vargaz: This should say 'incorrectly states' instead of 'lies'.

docs/advanced/runtime/docs/atomics-memory-model.md
```diff
@@ -0,0 +1,131 @@
...
+### `volatile.` prefix opcode in CIL
+
+When the `volatile.` prefix opcode is used in CIL, it imposes acquire/release semantics on the next non-prefix opcode.
+For loads, it results in acquire semantics. For stores, it results in release semantics.
+
+This prefix opcode has no effect on atomicity beyond the standard rules of the CLI.
+
+### `volatile` keyword in C\#
+
+The `volatile` keyword in C# compiles down to `Thread.VolatileRead` and `Thread.VolatileWrite` calls. Nothing more,
```
kumpera: mcs doesn't do that. Where did you get that from?

docs/advanced/runtime/docs/atomics-memory-model.md
```diff
@@ -0,0 +1,131 @@
...
+quantities on 32-bit systems, and the `Volatile` class did not exist when the `volatile` keyword was designed. Today,
+`volatile` on 64-bit quantities could conceivably be compiled down to `Volatile.Read` and `Volatile.Write` calls.
+
+### `Thread` class
+
+The `VolatileRead` and `VolatileWrite` methods perform loads and stores with acquire and release semantics,
+respectively. They guarantee absolutely nothing about atomicity beyond the standard rules of the CLI. In effect, this
+means that the 64-bit overloads of these methods are not atomic on 32-bit systems.
+
+There is a quirk in the Microsoft implementation where these methods actually use the `MemoryBarrier` method to insert
```
kumpera: Not microsoft, dotnet.

docs/advanced/runtime/docs/atomics-memory-model.md
```diff
@@ -0,0 +1,131 @@
...
+with a lock on 32-bit systems.
+
+The MSDN documentation incorrectly states that the C# compiler emits calls to this class's methods. It does not. This
+statement is probably intentional as, disregarding the implementation quirk with the `Thread` methods described above,
+C#'s `volatile` is actually supposed to behave the way that the methods on this class do. So the intention is likely to
+avoid having new code rely on the quirk in the `Thread` methods.
+
+### `Interlocked` class
+
+The methods on this class all have much the same semantics as those on the `Volatile` class, with the exception that
```
kumpera: I think this description is confusing. It's much simpler to say "The methods on this class are always atomic WRT to each other and those in the Volatile class. Method on this class impose sequential consistency."

We should state all cases in terms of atomicity and memory ordering.

alexrp:
@kumpera addressed all of your comments.

Btw, fun fact since you brought up the C# `volatile` thing.

From [`Thread.VolatileRead` documentation](https://msdn.microsoft.com/en-us/library/zdxzk5cw%28v=vs.110%29.aspx):

> In C#, using the volatile modifier on a field guarantees that all access to that field uses VolatileRead or VolatileWrite.

From [`Volatile` documentation](https://msdn.microsoft.com/en-us/library/system.threading.volatile%28v=vs.110%29.aspx):

> In C#, using the volatile modifier on a field guarantees that every access to that field uses the Volatile.Read and Volatile.Write methods

Neither statement is true. (╯°□°)╯︵ ┻━┻

txdv:
Is this page up? I can't find it ...

alexrp:
http://www.mono-project.com/docs/advanced/runtime/docs/atomics-memory-model/

PR #224, opened 2016-08-02
Add 'Lock-free GC Handles' post.

evincarofautumn:
A couple of things:
- I’m not sure when we want this post to actually go out, so presumably the date will have to change.
- Are we using a dialect of Markdown that supports tables?

### DIFF
--- a/_posts/2016-08-16-lock-free-gc-handles.md
+++ b/_posts/2016-08-16-lock-free-gc-handles.md
@@ -1,0 +1,180 @@
+---
+layout: blog
+title: "Lock-free GC Handles"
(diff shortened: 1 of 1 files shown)


akoeplinger:
> I’m not sure when we want this post to actually go out, so presumably the date will have to change.

Yeah. You could put the post into the [`_drafts`](https://jekyllrb.com/docs/drafts/) folder, but that'd only be useful locally as on GitHub we'd have to rename the file with the correct date anyway so it gets published. We can just do that before merging this PR as well.

> Are we using a dialect of Markdown that supports tables?

Yes.

PR #228, opened 2016-08-08
Packages list actualised

alexxn:
The following packages are not anymore in repository and/or changed their names:

mingw-runtime,mingw-binutils,mingw-gcc-core,mingw-gcc-g++,mingw-pthreads,mingw-w32api

Thus compilation is impossible if you try to follow official documentation from the website.
New versions proposed for update:

mingw64-i686-runtime, mingw64-i686-binutils, mingw64-i686-gcc-core, mingw64-i686-gcc-g++, mingw64-i686-pthreads, w32api-headers

that will preserve x32 versions of the packages


akoeplinger:
Thank you!

PR #232, opened 2016-08-31
remove my name from sqlite page

JoshData:
People email me from time to time with questions because they see my name on the Sqlite documentation page, but my involvement was literally more than a decade ago so I don't want to be a point of contact for this anymore!

Hello to anyone reading this that might remember me. :)


akoeplinger:
Sure thing :)

JoshData:
Thanks!

PR #234, opened 2016-09-03
Add profiler changes for 4.6.0.

alexrp:


docs/about-mono/releases/4.6.0.md
```diff
@@ -114,6 +114,102 @@ By doing so Mono runtime will not be able to output IL offsets instead of native
...
+version of the profiler to date. While there are a few new features, the focus
+has been on hardening the profiler to run on all sorts of platforms (including
+Linux, OS X, Android, and iOS) without crashing or otherwise destabilizing the
+program being profiled.
+
+#### New Features & Changes
+
+* The profiler now has support for gathering coverage data. This data can be
+  processed into XML by `mprof-report` and can then be consumed by
+  [MonoDevelop](http://www.monodevelop.com).
```
akoeplinger: Neat! Is this the `--coverage-out` option of mprof-report? It is not documented in the man page.
I also couldn't get it to work on a simple hello world, it just outputs an empty file.

How would I consume this XML in MD, just File->Open?

alexrp: I don't know much about how to actually use it (I didn't implement it). @iainx could you advise?

iainx: The XML can't be consumed by anything really.

@akoeplinger Does mprof-report give a summary of coverage data at the end of its output with your hello world test? the log profiler doesn't include coverage data by default, you'll need to enable it by passing the coverage option to the log profiler

akoeplinger: @iainx thx, I talked with alexrp on Slack and he also told me that coverage data needs to be explicitly enabled with `--profile=log:coverage` 😄 
The man page was updated in https://github.com/mono/mono/commit/bc8a54795e45d307008380f08340f2edec84704a to document this.

PR #236, opened 2016-09-06
Add first post in the profiler stability series.

alexrp:


alexrp:
cc @kumpera

akoeplinger:
Great post 😄 

Curious: did you run the change/PR through the benchmarking infrastructure as well? If yes, you could link to those results.

alexrp:
Yes, they are here: https://github.com/mono/mono/pull/2881#issuecomment-223415218

I'm not sure that linking to them in this post is a good idea though. I mean, I don't know for how long these results are kept around.

akoeplinger:
@alexrp links are going stale all the time on the Internet, it's not a unique problem to us and if they stay up for a reasonable time that's fine imho 😄

migueldeicaza:
Love it, can we commit, can we change the date to TOday, and commit?

lewurm:
@alexrp we do our best that those links work as long as possible :)

PR #242, opened 2016-10-05
Adding a post about static analysis

VasilievSerg:


txdv:
Isn't the post on your side enough?

VasilievSerg:
I think the readers would be interested to read about the errors found in the project. Perhaps, people will learn about static code analysis tools and realize how useful they may be. In the future we also would be glad to see a post of how these bugs were fixed.

PR #249, opened 2016-10-22
Add link to Slackware packages

gdsotirov:
I'm providing third party packages of Mono for [Slackware Linux](http://slackware/) (both 32 and 64 bit) from my site [SlackPack](http://sotirov-bg.net/slackpack/), so I hope they're useful for anyone trying to use Mono on this distribution.

### DIFF
--- a/download/index.html
+++ b/download/index.html
@@ -51,7 +51,6 @@ redirect_from:
             <li><a href="https://packages.gentoo.org/package/dev-lang/mono">Gentoo</a></li>
             <li><a href="https://software.opensuse.org/package/mono-complete">openSUSE</a></li>
             <li><a href="http://packages.ubuntu.com/mono-complete">Ubuntu</a></li>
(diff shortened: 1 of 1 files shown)


download/index.html
```diff
@@ -51,6 +51,7 @@
             <li><a href="https://packages.gentoo.org/package/dev-lang/mono">Gentoo</a></li>
             <li><a href="https://software.opensuse.org/package/mono-complete">openSUSE</a></li>
             <li><a href="http://packages.ubuntu.com/mono-complete">Ubuntu</a></li>
+            <li><a href="http://sotirov-bg.net/slackpack/search.cgi?q&equals;mono">Slackware</a></li>
```
akoeplinger: This section is for packages in the official package repositories of the various distributions, could you please move it down to "Community packages"?

Also no need to escape the = as `search.cgi?q=mono` will work fine (the current way doesn't work).

akoeplinger (changes requested):

akoeplinger (approved):

PR #250, opened 2016-10-26
Change some http links to https, especially download links

intgr:
There is no excuse to direct people to insecure download links when
https is available.

Changed links to websites:
- download.mono-project.com
- download.xamarin.com
- ftp.gnome.org
- ftp.gnu.org
- github.com (website is https-only)
- download.gnome.org (website is https-only)
- downloads.sourceforge.net (redirects browsers to https)
- packages.debian.org (website is https-only)

I didn't touch package manager links (Yum, APT, ...) because I presume
they are verified using digital signatures.


akoeplinger:
Fully agreed, thanks very much!

PR #254, opened 2016-11-27
mono_runtime_run_main crash if argc is zero

patricksuo:
env 
libmonosgen-2_0-1 x86_64  v4.6.2.7

```
#include <stdio.h>

#include <mono/jit/jit.h>
#include <mono/metadata/assembly.h>

int main(int argc, char **argv) {
	MonoDomain *domain = 0;
	MonoAssembly *assembly = 0;
	int retval = 0;

	domain = mono_jit_init("csharp_program.exe");
	assembly = mono_domain_assembly_open (domain, "csharp_program.exe");
	if (!assembly) {
		printf("%s", "mono_domain_assembly_open failed\n");
		return 1;
	}
	retval = mono_jit_exec (domain, assembly, argc, argv);
	printf("mono_jit_exec return with code %d\n", retval);

	return 0;
}
```

```
gcc main.c `pkg-config --cflags --libs mono-2`
./a.out
```
crash at https://github.com/mono/mono/blob/mono-4.6.0-branch/mono/metadata/object.c#L4090

### DIFF
--- a/docs/advanced/embedding/index.md
+++ b/docs/advanced/embedding/index.md
@@ -127,8 +127,14 @@ In the above example, the contents of \`file.exe' will be loaded into the domain
 To start executing code, you must invoke a method in the assembly, or if you have provided a static Main method (an entry point), you can use the convenience function:
 
 ``` c
-retval = mono_jit_exec (domain, assembly, argc, argv);
+retval = mono_jit_exec (domain, assembly, argc - 1, argv + 1);
 ```
+
+And execute c program like:
+``` bash
+./c_program file.exe
+```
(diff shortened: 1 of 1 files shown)


akoeplinger:
Found this old mailing list response where someone ran into the same problem: https://lists.dot.net/pipermail/mono-list/2009-November/043916.html

It seems the doc in this case assume your app is invoked like `./foo ./empty.exe`. I agree this is confusing, not sure if changing the example or simply adding a note is better. What do you think?

akoeplinger:
Thank you!

PR #255, opened 2016-12-16
Add cmake to the list of required packages.

scitesy:


akoeplinger:
For the record, cmake is only used for the new BoringSSL support in Mono 4.8+ right now.

PR #256, opened 2016-12-29
update

mejiajamie:
_site/
.vagrant
.jekyll-metadata


mejiajamie:
bad

akoeplinger:
We don't use WordPress so the .gitignore changes don't make sense. Closing this, thanks!

PR #259, opened 2017-02-15
Split Ubuntu and Debian install instructions

jesstelford:
Avoids confusion with libgdiplus causing "broken packages" on Ubuntu if accidentally added.

For example of the confusion, see [here](https://askubuntu.com/questions/729409/cannot-install-mono-depends-errors) (I also hit this bug and so wanted to contribute back to the docs).


directhex:
I'm gonna close this - the new (work in progress!) download page design subsumes this. See http://www.mono-project.com/download/alpha/

Those changes will move through beta & stable over the coming weeks - please help us make the new alpha page look great in the meantime!

PR #267, opened 2017-03-29
Refresh download instructions for new multi-repo world

directhex:
Plz don't merge until after d15.1 is stable, but I'd appreciate a review.

This obsoletes https://github.com/mono/website/pull/259


docs/getting-started/install/linux/index.md
```diff
@@ -87,6 +90,8 @@ The package ***referenceassemblies-pcl*** should be installed for PCL compilatio
 
 The package ***ca-certificates-mono*** should be installed to get SSL certificates for HTTPS connections. Install this package if you run into trouble making HTTPS connections.
 
+The package ***msbuild*** may be needed to compile very new projects which are not compatible with the deprecated MSBuild replacement, XBuild.
```
akoeplinger: This needs to be a dependency of mono-devel in the next release so I don't think we need to mention it explicitly.

docs/getting-started/install/linux/index.md
```diff
@@ -12,7 +12,7 @@ Regardless of your distribution, you will need the Mono Project GPG signing key,
 Debian, Ubuntu, and derivatives
```
akoeplinger: I'd like something similar to https://www.microsoft.com/net/core#linuxredhat where you choose your distro first so the page is smaller.

Should be easily doable given we already use such a switcher on https://github.com/mono/website/blob/gh-pages/download/index.html#L23, just horizontal.

docs/getting-started/install/linux/index.md
```diff
@@ -22,32 +22,46 @@ sudo apt-get update
...
+``` bash
+sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
+echo "deb http://download.mono-project.com/repo/debian jessie main" | sudo tee /etc/apt/sources.list.d/mono-xamarin.list
+sudo apt-get update
+```
+
+Run a package upgrade to upgrade existing packages to the latest available. Then install Mono as described in the [Usage](#usage) section.
 
-To enable mod_mono installation on Ubuntu 13.10 and later, and Debian 8.0 and later (and their derivatives), you will need to add a second repository to your system, **in addition** to the generic Debian/Ubuntu repository above (if you don't use sudo, be sure to switch to root):
+#### Ubuntu 14.04
```
akoeplinger: What's the status of Ubuntu 12.04? Can it continue to use the wheezy packages?

docs/getting-started/install/linux/index.md
```diff
@@ -22,32 +22,46 @@ sudo apt-get update
...
 
 ``` bash
-echo "deb http://download.mono-project.com/repo/debian wheezy-apache24-compat main" | sudo tee -a /etc/apt/sources.list.d/mono-xamarin.list
+sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
+echo "deb http://download.mono-project.com/repo/debian trusty main" | sudo tee /etc/apt/sources.list.d/mono-xamarin.list
```
akoeplinger: repo/debian is correct here, instead of e.g. repo/ubuntu ?

docs/getting-started/install/linux/index.md
```diff
@@ -63,17 +77,6 @@ Run a package upgrade to upgrade existing packages to the latest available. Then
 
 Users of CentOS or RHEL (or similar distributions) may need to add the [EPEL repository](https://fedoraproject.org/wiki/EPEL) to their system to satisfy all dependencies.
 
-openSUSE and SLES
```
akoeplinger: Does this mean we don't support openSUSE anymore?

docs/getting-started/install/linux/index.md
```diff
@@ -87,6 +90,8 @@ The package ***referenceassemblies-pcl*** should be installed for PCL compilatio
 
 The package ***ca-certificates-mono*** should be installed to get SSL certificates for HTTPS connections. Install this package if you run into trouble making HTTPS connections.
 
+The package ***msbuild*** may be needed to compile very new projects which are not compatible with the deprecated MSBuild replacement, XBuild.
+
 The module ***mono-xsp4*** should be installed for running ASP.NET applications.
 
 Notes
```
akoeplinger: can't comment on the line: should we do something about the "Security updates for older releases" section below? does it still apply in the new world?

docs/getting-started/install/linux/index.md
```diff
@@ -133,45 +138,116 @@ Alpha update channel
 
 If you want to test bleeding-edge updates to Mono, we now have an alpha update channel. Please report bugs promptly, so we have time to fix them before packages are migrated to stable!
 
-#### Debian, Ubuntu, and derivatives (alpha channel)
```
akoeplinger: we may want to consider splitting alpha/beta to their own page as it gets pretty long

docs/getting-started/install/linux/index.md
directhex: I hate how much we're coming to depend on MSbuild while its build process is so flakey :/

docs/getting-started/install/linux/index.md
directhex: Didn't even hit 1% of RPM users in the logs(!)

docs/getting-started/install/linux/index.md
directhex: Yeah, that indicates a directory on the server. Best I could do is symlink ubuntu to debian so it looks like it's different

docs/getting-started/install/linux/index.md
directhex: EOL in 28 days. Not worth thinking about any more.

docs/getting-started/install/linux/index.md
directhex: It's still accurate for the named releases. We could ditch the section if we're disclaiming responsibility for the versions listed, but would need to bring it back in future if any new vulns are found & we offer security updates to old versions.

docs/getting-started/install/linux/index.md
directhex: I agree.

docs/getting-started/install/linux/index.md
directhex: Sounds nice. Any ideas on implementation? Is there existing CSS to handle a horizontal switcher?

Actually, if we go that route, can we punt it to the main Download page?

akoeplinger: the line I linked to should have the skeleton that you need if you do `s/vertical/horizontal/`.

re: main Download page: I image it'd be too cramped there with the layout we have right now.

docs/getting-started/install/linux/index.md
akoeplinger: It's still the default on Travis iirc 😞

docs/getting-started/install/linux/index.md
akoeplinger: yeah a symlink sounds good to avoid any confusion.

docs/getting-started/install/linux/index.md
akoeplinger: Same story as with roslyn. Maybe we should do the same and just check in the binaries for now?

akoeplinger: Given that we need a story for people doing `./configure; make; make install` from the tarball that might be a good alternative because they'll be screwed otherwise once we fully deprecate xbuild (unless they somehow find out how to download+build msbuild)

docs/getting-started/install/linux/index.md
akoeplinger: Hmm, maybe we should just change this and _not_ provide symlinks for snapshots from e.g. 3.10 -> 3.10.X.X so users need to always pin to a specific version. A security fix would then just be "pin to a different version".

docs/getting-started/install/linux/index.md
directhex: Travis has 27 days to fix it before it becomes insecure, then

docs/getting-started/install/linux/index.md
directhex: OK, symlink in place. Also a symlink for alpha-wheezy -> alpha, so consistent structure is available. Should I change the docs to match?

docs/getting-started/install/linux/index.md
directhex: I'd like to give it a shot. I just need a modern jekyll installed to test a local gh-pages branch, right?

akoeplinger: Yes, just follow https://github.com/mono/website#setting-up-a-local-copy-of-the-website.

Newer Jekyll also supports `--incremental` so the whole site isn't rebuilt all the time which takes a while.

docs/getting-started/install/linux/index.md
akoeplinger: Yep, please change the docs to be consistent.

docs/getting-started/install/linux/index.md
```diff
@@ -133,45 +138,116 @@ Alpha update channel
...
+echo "deb http://download.mono-project.com/repo/debian beta-raspbianjessie main" | sudo tee /etc/apt/sources.list.d/mono-xamarin-beta.list
+sudo apt-get update
 ```
+
+#### CentOS, Fedora, and derivatives (beta channel)
+
+``` bash
+yum-config-manager --add-repo http://download.mono-project.com/repo/centos-beta/
+```
+
```
akoeplinger: we should keep the old version of the stable install page for the initial alpha release.

download/alpha.html
```diff
@@ -28,15 +28,63 @@
...
-          </ul>
+        <dl id="mono-download-linux" class="tabs horizontal" data-tab data-options="deep_linking: true; scroll_to_content: false">
+          <dd class="active"><a href="#download-lin-debian"><i class="fa fa-debian"></i><img src="/images/lin_debian.png" />&nbsp;Debian</a></dd>
+          <dd><a href="#download-lin-ubuntu"><i class="fa fa-ubuntu"></i><img src="/images/lin_ubuntu.png" />&nbsp;Ubuntu</a></dd>
+          <dd><a href="#download-lin-raspbian"><i class="fa fa-raspbian"></i><img src="/images/lin_raspbian.png" />&nbsp;Raspbian</a></dd>
+          <dd><a href="#download-lin-centos"><i class="fa fa-centos"></i><img src="/images/lin_centos.png" />&nbsp;CentOS</a></dd>
+        </dl>
+        <div class="tabs-content horizontal">
+          <div class="panel content active" id="download-lin-debian">
+            <h3>Mono is available as an Apt repository for Debian 8</h3>
```
akoeplinger: `Mono is available as an Apt repository for ` is redundant, remove.

download/alpha.html
```diff
@@ -28,15 +28,63 @@
...
-            <li><a href="/docs/getting-started/install/linux/#centos-fedora-and-derivatives-alpha-channel">CentOS, Fedora, and derivatives</a></li>
-          </ul>
+        <dl id="mono-download-linux" class="tabs horizontal" data-tab data-options="deep_linking: true; scroll_to_content: false">
+          <dd class="active"><a href="#download-lin-debian"><i class="fa fa-debian"></i><img src="/images/lin_debian.png" />&nbsp;Debian</a></dd>
+          <dd><a href="#download-lin-ubuntu"><i class="fa fa-ubuntu"></i><img src="/images/lin_ubuntu.png" />&nbsp;Ubuntu</a></dd>
+          <dd><a href="#download-lin-raspbian"><i class="fa fa-raspbian"></i><img src="/images/lin_raspbian.png" />&nbsp;Raspbian</a></dd>
+          <dd><a href="#download-lin-centos"><i class="fa fa-centos"></i><img src="/images/lin_centos.png" />&nbsp;CentOS</a></dd>
+        </dl>
+        <div class="tabs-content horizontal">
+          <div class="panel content active" id="download-lin-debian">
```
akoeplinger: when I visit the page it shows all distro instructions and not just the preselected Debian ones. Is that intended or a bug?

download/alpha.html
```diff
@@ -28,15 +28,63 @@
...
+            <h3>Mono is available as an Apt repository for Debian 8</h3>
+            <p>Add the Mono repository to your system:</p>
+            <pre><code class="language-bash">sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
+echo "deb http://download.mono-project.com/repo/debian alpha-jessie main" | sudo tee /etc/apt/sources.list.d/mono-xamarin-alpha.list
+sudo apt-get update</code></pre>
+            <p>Then update or install <strong><em>mono-devel</em></strong> to get started.</p>
+            <h3>Mono is available as an Apt repository for Debian 7</h3>
+            <p>Add the Mono repository to your system:</p>
+            <pre><code class="language-bash">sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
+echo "deb http://download.mono-project.com/repo/debian alpha-wheezy main" | sudo tee /etc/apt/sources.list.d/mono-xamarin-alpha.list
```
akoeplinger: I wonder if we could just make this `echo "deb http://download.mono-project.com/repo/debian alpha-$(lsb_release -cs) main" ... `

This would allows us to get rid of the repeated instructions. Do you know if lsb_release is available in most configurations (I heard something about some distros moving away from it)?

download/alpha.html
```diff
@@ -28,15 +28,63 @@
...
-          <p>These packages are built, tested and distributed by Xamarin. Use these if you want to use a stable, official and up-to-date version of Mono.</p>
-          <ul>
-            <li><a href="/docs/getting-started/install/linux/#debian-ubuntu-and-derivatives-alpha-channel">Debian, Ubuntu, and derivatives</a></li>
-            <li><a href="/docs/getting-started/install/linux/#centos-fedora-and-derivatives-alpha-channel">CentOS, Fedora, and derivatives</a></li>
-          </ul>
+        <dl id="mono-download-linux" class="tabs horizontal" data-tab data-options="deep_linking: true; scroll_to_content: false">
+          <dd class="active"><a href="#download-lin-debian"><i class="fa fa-debian"></i><img src="/images/lin_debian.png" />&nbsp;Debian</a></dd>
+          <dd><a href="#download-lin-ubuntu"><i class="fa fa-ubuntu"></i><img src="/images/lin_ubuntu.png" />&nbsp;Ubuntu</a></dd>
+          <dd><a href="#download-lin-raspbian"><i class="fa fa-raspbian"></i><img src="/images/lin_raspbian.png" />&nbsp;Raspbian</a></dd>
+          <dd><a href="#download-lin-centos"><i class="fa fa-centos"></i><img src="/images/lin_centos.png" />&nbsp;CentOS</a></dd>
```
akoeplinger: SVG instead of PNG (we can do it in a follow up cleanup if you don't have them handy).

akoeplinger (changes requested):

download/alpha.html
directhex: I don't have these handy. I wasn't sure how best to do these, since we use some magic Javascript to do the Win/Mac/Lin icons.

download/alpha.html
directhex: That's a bug. An interesting one!

The lack of pre-selection is (I think) due to a lack of the pre-selection JS in the alpha version of the page ( https://github.com/mono/website/blob/gh-pages/download/index.html#L108-L124 ) but showing all of them is definitely a bug

download/alpha.html
directhex: Pushed a change for this one

download/alpha.html
directhex: I don't believe it's guaranteed to exist on any distro, and it's a bit harder for people using deriavtives like Mint where they might know which Ubuntu or Debian they're equivalent to, but not how to modify the script. Maybe. Maybe that's a hypothetical user.

akoeplinger: Hmm, that's a good point. Let's keep it as it is for now.

Though we could strip the "Add the Mono repository to your system:" and "Then update or install mono-devel to get started" from each section and move it to a more general one. Which reminds me, what happened to the other info shown on the stable download page such as "Usage" and "Notes"?

directhex: How much of the Usage/Notes there is useful/interesting?

akoeplinger: Most of it? I think we can shorten it a bit, I can come up with something. We can do this after merging.

directhex: There's stuff that still needs fixing up, but what is blocking merging right now?

akoeplinger (approved):

PR #269, opened 2017-04-04
Document getenv change

alexanderkyte:


docs/about-mono/releases/5.0.0.md
```diff
@@ -40,5 +40,11 @@ code is shared between .NET Core and Mono and not just the Reference Source rele
 Bug Fixes
 =========
 
+Environment Memory Management
+-----------------------------
+As per the Unix specification, getenv may return a pointer into static memory. Future calls to getenv can overwrite
+the memory returned from previous calls to getenv. Mono kept many pointers into this memory while concurrently calling
+getenv on multiple threads. We now lock around the environment-changing functiosn and to duplicate all returned strings.
```
akoeplinger: typo: "functiosn and to duplicate all returned strings"

This fix doesn't have an effect for the user, correct? The other fix which caches env vars at app startup is coming later?

PR #290, opened 2017-05-19
Add compiler/build dependencies on Linux

luhenry:


docs/compiling-mono/linux/index.md
```diff
@@ -15,6 +15,22 @@ Building Mono is a very simple process as Mono has very few external dependencie
 
 On some distribution you will need to install some development tools which are not installed by default.
 
+Compiler/Build dependencies
+---------------------
+
+Mono depends on GCC 4.8+ or Clang 3.3+ (it needs to support C++11) and on CMake 2.8.10.
```
akoeplinger: We don't need cmake if you disable BTLS (at the moment, I guess this will change once your System.Native work lands right?)

docs/compiling-mono/linux/index.md
```diff
@@ -15,6 +15,22 @@ Building Mono is a very simple process as Mono has very few external dependencie
...
+
+| Distribution | GCC         | Clang   | CMake      |
+| ------------ | ----------- | ------- | ---------- |
+| Debian 7     | ~~4.7/4.6~~ | ~~3.0~~ | ~~2.8.9~~  |
```
akoeplinger: not sure I understand what the crossed-out versions mean? Or more generally, what is the purpose of this table?

docs/compiling-mono/linux/index.md
luhenry: The C++11 need is for CoreFX, so we will need CMake

docs/compiling-mono/linux/index.md
luhenry: The cross means it doesn't satisfy the version needed. The purpose of this table is more of a note to help us track which version is available on which platform, so if we ever need to change our requirements, we know what platform we would break.

akoeplinger: Ok, can you add a short comment above/below the table to that regard?

PR #293, opened 2017-06-10
Updated to HTML5 standard & added specific meta tags

ghost:
I removed the "/" from the link tags as it does not follow the HTML5 standard and added a couple of meta tags to improve experience on certain devices and even though not that important for SEO reasons.

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -2,19 +2,26 @@
 <!--[if IE 9]><html class="lt-ie10" lang="en"><![endif]-->
 <html class="no-js" lang="en">
   <head>
-    <meta charset="utf-8" />
-    <meta name="viewport" content="width=device-width, minimum-scale=1.0, maximum-scale=1.0" />
-    <meta http-equiv="X-UA-Compatible" content="IE=Edge" />
+    <meta name="viewport" content="width=device-width, minimum-scale=1.0, maximum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="IE=Edge">
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+    <meta http-equiv="cleartype" content="on">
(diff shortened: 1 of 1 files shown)


akoeplinger:
Hey, thanks for the PR!

Unfortunately it looks like none of the changes provide a benefit for us, here are the reasons:

a) The HTML5 standard [allows](https://www.w3.org/TR/html5/syntax.html#start-tags) `/>` on void elements (and my mind prefers properly closed XML tags):

> 6. Then, if the element is one of the void elements, or if the element is a foreign element, then there may be a single "/" (U+002F) character. This character has no effect on void elements, but on foreign elements it marks the start tag as self-closing.

> Void elements
area, base, br, col, embed, hr, img, input, keygen, link, meta, param, source, track, wbr

b) The `<meta charset="utf-8" />` should come as early as possible so the browser can decide about the charset and doesn't need to reparse the document so we can't remove it.

c) `<meta http-equiv="cleartype" content="on">` seems to be recommended against since it only applies to very old mobile/desktop IE which we don't care about

d) Similar reason for `<meta name="HandheldFriendly" content="True">`, seems to be only used by some very old mobile browsers.

e) `<meta name="apple-mobile-web-app-capable"` and `apple-mobile-web-app-title` -> we're a website, not a web app.

f) `<meta http-equiv="Content-Type" content="text/html; charset=utf-8">` -> this is already sent by the webserver, no need to specify it again and waste bytes

g) `<meta name="identifier-url"` -> can't find too much information about it, I don't think it's very useful.

h) `<link rel="canonical" href="http://www.mono-project.com/">` and `<meta name="url"  content="http://www.mono-project.com/"` this needs to point to a specific page to be useful. We'd be pointing to the root page on every page since this file is a template, so this would actually be harmful.

Thanks again for your time, but I'll rather not take these changes.

PR #297, opened 2017-07-06
Update repl.md

MichaelMcGlothlin:
Clarified that when loading C# script files the filenames should end in .cs as it appears that .csx and other possible options do not work.


akoeplinger:
Indeed: [repl.cs#L379-L380](https://github.com/mono/mono/blob/b0057f2259acc7f0716b36346d739dc1bdf7afe9/mcs/tools/csharp/repl.cs#L379-L380). Thanks!

PR #300, opened 2017-07-20
Updating the Boo documentation page

masonwheeler:
Improving description and feature set
Replacing broken Codehaus links with the modern GitHub versions

### DIFF
--- a/archived/boo.md
+++ b/archived/boo.md
@@ -1,6 +1,6 @@
 ---
 title: "Boo"
-lastmodified: '2005-09-21'
+lastmodified: '2017-07-20'
 redirect_from:
   - /Boo/
 ---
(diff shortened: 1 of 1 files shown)


masonwheeler:
Thanks.  But that didn't remove the "this page is likely outdated" notice, even though I updated the date metadata.  How do we get this in the "new documentation" section?

akoeplinger:
@masonwheeler yeah, that's because it's in the `archived/` folder which is basically stuff that we don't really care about anymore (I did a categorization when the site was moved from the old MediaWiki based site to GitHub three years ago).

Boo is a third party project and not really related to Mono so I thought it doesn't make too much sense to keep this page on the Mono website but feel free to prove me wrong 😄

PR #307, opened 2017-08-11
Update the articles about Clang's sanitisers

cherusker:
- add an article about ASan
- add information about the `Unlocked* ()` functions to the TSan article
- put the information about blacklists and flags into separate articles
- improve the general structure of the Clang sanitisers documentation

### DIFF
--- a/docs/debug+profile/clang/addresssanitizer.md
+++ b/docs/debug+profile/clang/addresssanitizer.md
@@ -1,0 +1,38 @@
+---
+title: AddressSanitizer (Clang Sanitizers)
+---
+
+The [AddressSanitizer](https://clang.llvm.org/docs/AddressSanitizer.html) is integrated into [Clang](https://clang.llvm.org/) 3.4+ and therefore, if Clang is installed, using it with Mono's `autogen.sh` is as simple as running the following:
+
+``` bash
+$ cd /home/root/of/mono
+$ ./autogen.sh --prefix=/usr/local CC=clang CFLAGS=-fsanitize=address LDFLAGS=-fsanitize=address CXX=clang++
+$ make
+...
+```
+
+Please keep in mind that `CXX` has to be set to `clang++` to avoid unwanted side effects and compiling errors.
+
+After `autogen.sh` has been executed as suggested above, the following `make` command uses Clang and injects necessary reporting functions.
+
+In the context of Mono, the AddressSanitizer (without the LeakSanitizer, [see below](#ignore-the-leaks)) produces a lot less false positives than the [ThreadSanitizer](/docs/debug+profile/clang/threadsanitizer/), which makes working with it more straight forward.
+
+Dynamic Blacklisting
+--------------------
+
+Still, false positives need to be hidden, and so far, the best option is to work with blacklists. Please have a look at [this article](/docs/debug+profile/clang/blacklists/) to find out how to work with blacklists in combination with Mono.
+
+Since there are only a handful of false positives, no macros or special functions can be used to permanently blacklist anything directly within the code (August 2017).
+
+Runtime Options
+---------------
+
+Runtime options can be provided by using the environment variable `ASAN_OPTIONS`. Please find detailed information about runtime options and Mono in [this article](/docs/debug+profile/clang/customisation/#runtime-options) and all available options on [Google's GitHub page](https://github.com/google/sanitizers/wiki/AddressSanitizerFlags#run-time-flags). The following two options prove to be particularly useful:
+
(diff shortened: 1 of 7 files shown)


docs/debug+profile/clang/blacklists.md
```diff
@@ -0,0 +1,51 @@
...
+
+``` bash
+$ touch /home/some/path/blacklist
+$ cd /home/root/of/mono
+$ ./autogen.sh --prefix=/usr/local CC=clang CFLAGS="-fsanitize=thread -fsanitize-blacklist=/home/some/path/blacklist" LDFLAGS=-fsanitize=thread CXX=clang++
+$ make
+...
+```
+
+Currently (August 2017), it seems like a good idea to leave all blacklist files outside of folders that contain `Makefile` files (`mono/eglib/`, `mono/metadata/`, ...), as sometimes `autogen.sh` treats blacklist files like source files and lists them as dependencies. As an additional and positive side effect, blacklist files will not interfere with git if they are hosted outside of the repository.
```
akoeplinger: that's interesting, not sure how this could happen?

docs/debug+profile/clang/addresssanitizer.md
```diff
@@ -0,0 +1,38 @@
+---
+title: AddressSanitizer (Clang Sanitisers)
```
akoeplinger: `s/Sanitisers/Sanitizers` for consistent spelling (applies to all articles).

docs/debug+profile/clang/addresssanitizer.md
```diff
@@ -0,0 +1,38 @@
...
+``` bash
+$ cd /home/root/of/mono
+$ ./autogen.sh --prefix=/usr/local CC=clang CFLAGS=-fsanitize=address LDFLAGS=-fsanitize=address CXX=clang++
+$ make
+...
+```
+
+Please keep in mind that `CXX` has to be set to `clang++` to avoid unwanted side effects and compiling errors.
+
+After `autogen.sh` has been executed as suggested above, the following `make` command uses Clang and injects necessary reporting functions. Later, `make` uses [roslyn](https://github.com/dotnet/roslyn) to compile `mcs` with the freshly built Mono runtime.
```
akoeplinger: I'd remove the last sentence, it doesn't really add anything valuable (and could be even a bit confusing since Mono's C# compiler is also called `mcs`).

docs/debug+profile/clang/addresssanitizer.md
```diff
@@ -0,0 +1,38 @@
...
+Since there are only a handful of false positives, no macros or special functions can be used to permanently blacklist anything directly within the code (August 2017).
+
+Runtime Options
+---------------
+
+Runtime options can be provided by using the environment variable `ASAN_OPTIONS`. Please find detailed information about runtime options and Mono in [this article](/docs/debug+profile/clang/customisation/#runtime-options) and all available options on [Google's GitHub page](https://github.com/google/sanitizers/wiki/AddressSanitizerFlags#run-time-flags). The following two options prove to be particularly useful:
+
+### Ignore the Leaks
+
+By default, the AddressSanitizer starts the [LeakSanitizer](https://clang.llvm.org/docs/LeakSanitizer.html) as soon the program terminates otherwise successfully. In the context of Mono this means several 100 (sometimes up to a few 1000) reported leaks. `detect_leaks=0` tells the AddressSanitizer to shut down without starting the LeakSanitizer.
```
akoeplinger: typo: as soon _as_ the program terminates

docs/debug+profile/clang/blacklists.md
cherusker: I didn't change that part much (had it there with the first version of the TSan article already); it was interesting but running `./autogen` and `make` consecutively with and without TSan options and having the blacklist somewhere inside the `mono/` folder (and not) gave more trouble (something about missing files) than having it somewhere outside (somehow it seemed to be connected to the built libraries inside the `./libs` folders). We can remove this part however, until I know exactly what's going on there.

docs/debug+profile/clang/addresssanitizer.md
cherusker: thanks!

docs/debug+profile/clang/addresssanitizer.md
cherusker: I removed the last sentence in the AddressSanitizer's article and slightly altered it in the ThreadSanitizer's article (mentioning the `mscorelib.dll` directly instead of the whole `mcs/` folder)

akoeplinger (approved):

PR #318, opened 2017-08-31
Added the report to CSCache / Cache tool for mono (Calancea Daniel)

dcalance:


akoeplinger:
Thanks for the report!

Issue #343, opened 2018-02-06
Broken link in Runtime documentation


loic-sharma:
Under [the `Compilation Engine` section of the Runtime's documentation](https://github.com/mono/website/blob/gh-pages/docs/advanced/runtime/index.md#compilation-engine), the link "http://primates.ximian.com/~lupus/slides/jit" is broken.

akoeplinger:
There's a copy on archive.org: http://web.archive.org/web/20070226130104/http://primates.ximian.com:80/~lupus/slides/jit/index.html

Though given that these slides were written in 2002 I wonder how relevant they are today.

loic-sharma:
Hm, you're right, it's probably not very relevant. Really, I want to check out Mono's JIT compiler so I can see if it supports WebAssembly :)

akoeplinger:
It's probably better to ask those questions on the main Mono repo (https://github.com/mono/mono/issues) or on [Gitter](https://gitter.im/mono/mono) :)

PR #344, opened 2018-02-08
[runtime] backup of Paolo's JIT slides

lewurm:
I think they are worth archiving. See https://github.com/mono/website/issues/343#issuecomment-363622964


akoeplinger:
Awesome, I briefly considered doing this but then postponed it :)

I ran the images through ImageOptim: Saved 53 KB out of 437 KB. 12,4 % per file on average

PR #345, opened 2018-02-09
Update aspnet.md

jaszb:


docs/web/aspnet.md
```diff
@@ -6,7 +6,7 @@ redirect_from:
   - /XSP/
 ---
 
-Mono has an implementation of ASP.NET 2.0, ASP.NET MVC and ASP.NET AJAX.
+Mono has an implementation of ASP.NET Core 2.0, ASP.NET MVC and ASP.NET AJAX.
```
akoeplinger: that's not correct, we don't have an implementation of ASP.NET Core in Mono.

We do ship an implementation of traditional ASP.NET, I'd suggest just removing the `2.0` part (it should be up to 4.something at the moment)

akoeplinger (changes requested):

PR #350, opened 2018-02-15
Add Fedora download instructions

harry-cpp:
cc. @directhex 

Notes:
- Should I add a note somewhere stating that it is using CentOS repo?
- ~~I snuck in a `mono-` prefix to the CentOS repo output~~


akoeplinger:
What if we combine them, i.e. call the tab `CentOS + Fedora` since they're using the same repo?

harry-cpp:
> What if we combine them, i.e. call the tab CentOS + Fedora since they're using the same repo?

CentOS uses yum, Fedora uses dnf, they are a bit different.

akoeplinger:
Thank you!

harry-cpp:
Thank you for merging as well.

PS. Fedora has `sudo` installed by default so you could have kept the command, tho the changed command is fine as well.

akoeplinger:
It doesn't have sudo in the Docker container so I figured I'd make it consistent with the CentOS instructions :)

PR #353, opened 2018-02-27
Explicitly enumerate supported Linux build architectures

directhex:


_includes/download.html
```diff
@@ -260,7 +260,7 @@ <h2>
             </h2>
             <p>The package repository hosts the packages you need, add it with the following commands <b>in a root shell</b>.</p>
             <p><i>Note:</i> the packages should work on newer CentOS versions too but we only test the ones listed below.</p>
-            <h3>CentOS 7</h3>
+            <h3>CentOS 7 (x86_64)</h3>
```
akoeplinger, line 263: I assume this is what's it called in the RPM world and we don't want to call it `amd64` ?

directhex: Aye.

`mono-core-5.10.0.140-0.xamarin.2.epel6.x86_64.rpm`
`mono-core-5.10.0.140-0.xamarin.2.epel6.i686.rpm`

akoeplinger (approved):

PR #355, opened 2018-02-28
Fix typo

saraedum:
I am not sure why we are explaining the case of OpenSUSE here, but I am quite sure that they do not call it `hdtocs` in SUSE.


akoeplinger (approved):
Thank you!

PR #356, opened 2018-02-28
Dead link

saraedum:
The configuration tool only produces a 503 for me. I am not sure whether that tool is dead or just temporarily gone. If it is gone for good, we should remove that section.


akoeplinger (approved):
Yeah it's dead, thanks!

PR #357, opened 2018-02-28
Minor grammar fix

kas:
Hello:

As I was reading about the companies using Mono, I noticed an extra `a` in the message found at the end of the page.


luhenry:
Thanks for the contribution! :)

PR #377, opened 2018-04-23
Add LLVM instructions for nightly .debs

directhex:


akoeplinger (approved):

marek-safar:
Why for nightlies only?

directhex:
@marek-safar Because right now only nightlies are compiled against LLVM, until other issues like https://github.com/mono/mono/issues/8410 are fixed

PR #386, opened 2018-05-29
Update mono-basics.md

ashirafumiiro:
csc replaced with mcs because csc is not a recognised command in kali linux after installing mono

### DIFF
--- a/docs/getting-started/mono-basics.md
+++ b/docs/getting-started/mono-basics.md
@@ -25,7 +25,7 @@ public class HelloWorld
 
 To compile, use csc:
 
-    csc hello.cs
+    mcs hello.cs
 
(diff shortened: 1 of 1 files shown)


marek-safar:
What Mono version are you using?

ashirafumiiro:
Am using version 4.6.2 from the kali Linux repository and I have got the
reason why.
Am going to update my version.
Thank you!

lextm:
@ashirafumiiro you should close this PR.

luhenry:
`csc` is the way forward, we are maintaining `mcs` but it's not unlikely to be lagging behind in terms of features. The documentation is mostly for the latest version, so updating it to use `mcs` doesn't make sense.

PR #403, opened 2018-10-01
Create 403.html

Kundan28:


akoeplinger:
I'm not sure I understand why this is useful, can you please explain?

Kundan28:
@akoeplinger We can make changes to the 403.html as we want, and don't have to see the default version that pops up. Now the page is simple but it can be made more attractive.

akoeplinger:
@Kundan28 I still don't understand, we already have a 404.html: https://github.com/mono/website/blob/gh-pages/404.html

What special handling does the 403.html page you added get?

Kundan28:
There was typo in my comment, I was talking about 403.html

akoeplinger:
Ok, as I said we already have a 404.html so if you'd like to make improvements to it please edit that file in a new PR :)

Kundan28:
@akoeplinger I am talking about 403.html, why are you talking about 404.html

akoeplinger:
@Kundan28 ok, but I still don't see what adding this file accomplishes? It seems to get no special treatment from GitHub Pages.

Kundan28:
The link that you shared above also not contain 403.html

akoeplinger:
yes, I initially thought you were talking about 404.html - still my question stands: why should we add 403.html?

PR #404, opened 2018-10-04
Mention how to run csc on Linux

aleb:
On https://packages.debian.org/stretch/all/mono-devel/filelist it can be seen it's mono-csc

### DIFF
--- a/docs/getting-started/mono-basics.md
+++ b/docs/getting-started/mono-basics.md
@@ -23,7 +23,7 @@ public class HelloWorld
 }
 ```
 
-To compile, use csc:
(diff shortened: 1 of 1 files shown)


akoeplinger:
Hey, that is an old symlink, introduced to make it easier to rebuild the archive against newer c# compiler versions, when there was mcs, gmcs, dmcs, smcs, etc.

The Mono version in Debian's package repositories (4.6.2.7) is pretty old. The current Mono from https://www.mono-project.com/download/stable/ installs a normal `csc` command so our docs are correct, thus I'll close the PR. Thanks :)

PR #406, opened 2018-10-07
Better Aligned text in terms of UI

tannmaysgupta:
The new update makes the website semantically better and also adds a article tag which is a better tag in terms of ease of access for browsers and makes it a inclusive design.

### DIFF
--- a/404.html
+++ b/404.html
@@ -5,8 +5,8 @@ permalink: /404.html
 ---
 <div class="row">
   <div class="small-12 columns">
-     <h1 style="text-align: center;">We could not find that page</h1>
-     <article style="text-align: center;">
+     <h1>We could not find that page</h1>
+     <article>
      Please check that the link you used is correct.
      </article>
   </div>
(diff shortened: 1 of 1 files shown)


404.html
```diff
@@ -5,9 +5,9 @@
 ---
 <div class="row">
   <div class="small-12 columns">
-     <h1>We could not find that page</h1>
-     <p>
+     <h1 style="text-align: center;">We could not find that page</h1>
+     <article style="text-align: center;">
```
akoeplinger: All our other pages use left-aligned text so I'd like to keep it that way here too, can you please remove that part?

I'm fine with the p -> article change.

akoeplinger (changes requested):

404.html
tannmaysgupta: yeah sure I'll make the changes back to left align

akoeplinger (approved):

PR #407, opened 2018-10-09
full width nav, spacing in hero section

haych8:


_layouts/base.html
```diff
@@ -45,7 +45,7 @@
   </head>
   <body>
     <div id="main-wrapper">
-      <div id="header-nav" class="contain-to-grid">
+      <div id="header-nav" class="container-fluid">
```
akoeplinger, line 48: `container-fluid` doesn't exist, it's probably the same effect as removing it.

css/main.css
```diff
@@ -170,19 +170,19 @@ footer h6 a {
...
-  margin-bottom: 15px;
+  margin: 15px 0 40px;
   font-weight: bold;
 }
 
 #index-teaser {
   background: #EBEBEB;
   color: #222;
-  margin-top: 20px;
+  margin-top: 100px;
```
akoeplinger, line 185: these values are way too large, on a screen with less vertical space all you're seeing now is just the hero section.

akoeplinger (commented):
Thanks for your effort but I don't think I'd like to take this change. What we have served us well so far :)

PR #410, opened 2018-10-23
Change the font

Leticiasb:
I believe this new font is more modern and better represent the project


akoeplinger:
Thanks but we're not interested in changing the header font at this time :)

PR #412, opened 2018-10-26
Added a color transition on the header links

vaibhav111tandon:


akoeplinger:
Thanks but this isn't something that fits with the current design of the website :)

PR #418, opened 2018-11-02
small typo when trying to compile mono in terminal

alsidneio:
had trouble when trying to follow the documentation directions.  running "csc hello.cs" returns the following: 
` csc hello.cs

Command 'csc' not found, but can be installed with:

sudo apt install chicken-bin`

However when I ran the "msc hello.cs" it created the executable. This on a linux machine: debian distro

### DIFF
--- a/docs/getting-started/mono-basics.md
+++ b/docs/getting-started/mono-basics.md
@@ -23,9 +23,9 @@ public class HelloWorld
 }
 ```
 
-To compile, use csc:
+To compile, use msc:
 
-    csc hello.cs
+    msc hello.cs
 
 The compiler will create "hello.exe", which you can run using:
 


akoeplinger:
`csc` is the C# compiler, you'll need it for a lot of the build-related parts of Mono to work.

@directhex do we actually not ship `csc` in our packages? I thought that only applied to the Debian repo.

directhex:
@akoeplinger we ship `csc` in mono-project.com packages, although this will cause errors for users with `chicken-bin` already installed. I'm not sure how the `file-not-found` apt integration gets its list of proposed packages, though, so we might be missing from there.

akoeplinger:
Ok so I assume @alsidneio didn't install the packages from mono-project.com then but the Debian repo. Closing the PR then since it wouldn't happen with our packages.

PR #431, opened 2019-02-08
Corrected reference to tlstest (mono-basics.md) which is not working in master

sorin-postelnicu:
Since tlstest was removed in master, I replaced the link to tlstest with an older branch.
I have also corrected the command to compile it, because it needs a reference to Mono.Security.dll otherwise we get this error: 
tlstest.cs(23,21): error CS0234: The type or namespace name 'Protocol' does not exist in the namespace 'Mono.Security'. Are you missing an assembly reference?

### DIFF
--- a/docs/getting-started/mono-basics.md
+++ b/docs/getting-started/mono-basics.md
@@ -40,10 +40,10 @@ Hello Mono World
 HTTPS connections
 -----------------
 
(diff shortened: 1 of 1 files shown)


akoeplinger:
Thank you for the contribution!

We removed tlstest on purpose since it only works with the legacy TLS provider.

I replaced the mention on the website with a better check here: https://github.com/mono/website/commit/0d559d4a521d543544762b05ea47cbd4c459a153

PR #435, opened 2019-03-06
versioning.md: Add Roslyn version to table

sandyarmstrong:
These values have been determined by looking at the value of `CSC_LOCATION` in the `configure.ac` of each branch.


sandyarmstrong:
~@marek-safar I'm a little concerned that PRs like https://github.com/mono/mono/pull/11573 have merged (and been backported all the way to `2018-08`), seeming to indicate that we've moved to Roslyn 2.9, but `CSC_LOCATION` has not been updated. Is that intended?~

EDIT: After asking @akoeplinger, it seems the intent of that PR was to update `csc-dim`, not `csc`. Please disregard this question. :-)

akoeplinger (approved):

PR #437, opened 2019-03-20
Update mono-basics.md

Teklie:
I have changed [static public void Main ()] to [public static void Main(string[] args)]


marek-safar:
Thank you!

PR #441, opened 2019-04-19
Describe how to use LLDB to debug Mono binaries

lambdageek:
You can use LLDB for source-level debugging of Mono.  All this stuff is on the internet in various places, but let's all bring it together into one document.


lewurm:
this is awesome, thank you for writing it down 🙂

PR #442, opened 2019-05-01
Add llvm optimization blog post

alexanderkyte:


alexanderkyte:
@Therzok suggested I have an LLVM IR before/after example program. I think that the nonnull-motivating benchmark is a good one for that. I'll get some IR from an old mono at some point.

rondales (approved):

PR #456, opened 2019-08-18
page mono-basics, change compiler csc to msc 

johannesmik:
In the Hello World example of the [Mono Basics](https://www.mono-project.com/docs/getting-started/mono-basics/) page, the csc compiler is used to compile the example.

This pull request suggests that instead of the csc compiler, the mcs compiler should be used, together with a link to the msc description page https://www.mono-project.com/docs/about-mono/languages/csharp/

### Background

I installed `mono-complete` on a machine running Ubuntu, and tried to follow the Hello World example to verify if everything was installed fine. However `csc` was not installed, so the instructions didn't work. Took me a few minutes to get to know that it should be `mcs` instead.

### DIFF
--- a/docs/getting-started/mono-basics.md
+++ b/docs/getting-started/mono-basics.md
@@ -24,9 +24,9 @@ public class HelloWorld
 }
 ```
 
-To compile, use [msc](/docs/about-mono/languages/csharp/):
+To compile, use [mcs](/docs/about-mono/languages/csharp/):
 
-    msc hello.cs
+    mcs hello.cs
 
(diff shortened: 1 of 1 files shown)


lextm:
Please reveal your Ubuntu version and Mono version. I believe the documentation should cover only the latest Mono, so your suggested changes to use `mcs` is far obsolete.

johannesmik:
Sorry for the confusion caused.

I am on Ubuntu 18, I downloaded the mono version available in the official sources (which is mono 4.6).

I didn't know that since mono 5.0, the csc compiler became part of mono.

PR #458, opened 2019-09-03
csc to msc

hatzhang:
totally new to this. but csc not found and mono installed on my ubuntu,while msc works.

### DIFF
--- a/docs/getting-started/mono-basics.md
+++ b/docs/getting-started/mono-basics.md
@@ -24,9 +24,9 @@ public class HelloWorld
 }
 ```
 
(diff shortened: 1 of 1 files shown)


marek-safar:
@directhex does csc come with the different package?

directhex:
The packages in Ubuntu don't come with csc available at all, since Roslyn isn't really buildable from source right now. Our packages on mono-project.com have Roslyn in the `mono-roslyn` package, which is installed by `mono-devel`. In any case, `mcs` not `msc`

PR #468, opened 2020-01-18
fix: List of supported AOT platforms

bruno-garcia:
The `limitation: Platform` list is out of sync with the section below `Supported Platforms`

Perhaps it would be worth removing this section altogether? Since it's again listed a few lines below?


akoeplinger:
The section below is for the "Generic ValueType Sharing" feature. I'm not sure if we have platforms where we support FullAOT but not gsharedvt though.

akoeplinger (approved):

PR #478, opened 2020-02-10
Reflect deprecation of IOMAP for Mono 6.0.0 upwards

WizzardMaker:
The IOMAP feature is now deprecated, but the docs still say, that you can use it!


akoeplinger (approved):
Thank you!

PR #482, opened 2020-03-12
Removed advice to change autocrlf setting globally

LunicLynx:
The autocrlf setting should not be `input` on windows. And if it is explicity required for mono, it should only be set for the project, not globally.
It would be best to set it via `.gitattributes`. 
https://git-scm.com/docs/gitattributes


akoeplinger (approved):
Yeah I think we shouldn't need this anymore these days. Thanks!

PR #494, opened 2020-08-24
Add post on native library loading in .NET 5

CoffeeFlux:


_posts/2020-08-24-native-loader-net5.md
```diff
@@ -0,0 +1,38 @@
...
+
+A more subtle, yet no less impactful change is that native library loading now defaults to `RTLD_LOCAL` to be consistent with CoreCLR and Windows, as opposed to our historical behavior of `RTLD_GLOBAL`. This change prevents symbol collision, and will both break and enable various scenarios/libraries. What this means in practice is that on Unix-like platforms, libraries are longer loaded into a single global namespace and when looking up symbols, the specified library must be correctly specified.
```
Therzok:
```suggestion
A more subtle, yet no less impactful change is that native library loading now defaults to `RTLD_LOCAL` to be consistent with CoreCLR and Windows, as opposed to our historical behavior of `RTLD_GLOBAL`. This change prevents symbol collision, and will both break and enable various scenarios/libraries. What this means in practice is that on Unix-like platforms, libraries are longer no loaded into a single global namespace and when looking up symbols, the specified library must be correctly specified.
```

_posts/2020-08-24-native-loader-net5.md
```diff
@@ -0,0 +1,38 @@
...
+
+Historically, in Mono on Linux it was possible to load library `foo` containing symbol `bar`, and then invoke `bar` with a P/Invoke like so: 
+```
+[DllImport("asdf")]
+public static extern int bar();
+```
+
+This will no longer work. For that P/Invoke to function correctly, the attribute would need to use the correct library name: `[DllImport("foo")]`. A lot of code in the wild that was using incorrect library names will need to be updated. However, this means that when loading two libraries containing the same symbol name, there is no longer a conflict.
+
+There have been some embedding API changes as part of this. `MONO_DL_MASK` is no longer a full mask, as `MONO_DL_GLOBAL` has been introduced to specify `RTLD_GLOBAL`. If both `MONO_DL_LOCAL` and `MONO_DL_GLOBAL`, are set, Mono will use local. See mono/utils/mono-dl-fallback.h for more info.
```
Therzok, line 46: Interesting choice here. Why not imply the absence of `GLOBAL` to be local? That way there would be no way to specify both.

CoffeeFlux: Do you mean the absence of `LOCAL`? The embedding API historically has defaulted to `GLOBAL`, but that's what I've changed here; it now defaults to `LOCAL`. If I don't extend it, there will be no way to override that.

CoffeeFlux: For what it's worth, I considered leaving the embedding API to default to `GLOBAL` and just changing the P/Invoke behavior, but ultimately decided this would be oddly confusing/inconsistent and not push people towards just using `LOCAL` by default.

_posts/2020-08-24-native-loader-net5.md
```diff
@@ -0,0 +1,38 @@
...
+
+One area with multiple highly impactful changes to the runtime internals is library loading. For managed assemblies, Mono now follows the algorithms outlined on [this page](https://docs.microsoft.com/en-us/dotnet/core/dependency-loading/loading-managed), which result from the removal of `AppDomain`s and the new `AssemblyLoadContext` APIs. The only exception to this is that Mono still supports bundles registered via the embedding API, and so the runtime will check that after the active `AssemblyLoadContext` cache.
+
+The managed loading changes are fairly clear and well documented, but unmanaged library loading has changed in numerous ways, some of them far more subtle.
+
+First off, Mono now follows [this algorithm](https://docs.microsoft.com/en-us/dotnet/core/dependency-loading/loading-unmanaged) for P/Invoke resolution. Notably absent here is DllMap's XML config files, which have been disabled entirely in .NET 5. The DllMap embedding APIs are also disabled by default on desktop platforms, but can be enabled by building with `--enable-minimal=dllmap`.
```
lambdageek: @CoffeeFlux This doesn't look right. I don't think `--enable-minimal=dllmap` will *bring back* dllmap support.

`--enable-minimal=foo` usually _disables_ support for feature *foo* by defining `DISABLE_FOO`.   And as far as I know there's no `configure` thing to turn on a disabled minimal feature that's off by default.

Looking at `configure.ac` I think the only way to turn dllmap back on is to hack the .ac file or not to build with the `netcore` preset.

I think it might be better to just leave it at "We dropped support for dllmap in net5".

_posts/2020-08-24-native-loader-net5.md
```diff
@@ -0,0 +1,38 @@
...
+
+A more subtle, yet no less impactful change is that native library loading now defaults to `RTLD_LOCAL` to be consistent with CoreCLR and Windows, as opposed to our historical behavior of `RTLD_GLOBAL`. This change prevents symbol collision, and will both break and enable various scenarios/libraries. What this means in practice is that on Unix-like platforms, libraries are no longer loaded into a single global namespace and when looking up symbols, the library must be correctly specified.
```
lambdageek: I think it would be good to explain what `RTLD_LOCAL`/`RTLD_GLOBAL` do (or just link to some manpages).

_posts/2020-08-24-native-loader-net5.md
```diff
@@ -0,0 +1,38 @@
...
+
+Historically, in Mono on Linux it was possible to load library `foo` containing symbol `bar`, and then invoke `bar` with a P/Invoke like so: 
+```
+[DllImport("asdf")]
```
lambdageek, line 40: nit: add `// note the incorrect library name` on this line to highlight the problem

_posts/2020-08-24-native-loader-net5.md
```diff
@@ -0,0 +1,38 @@
...
+
+There have been some embedding API changes as part of this. `MONO_DL_MASK` is no longer a full mask, as `MONO_DL_GLOBAL` has been introduced to specify `RTLD_GLOBAL`. If both `MONO_DL_LOCAL` and `MONO_DL_GLOBAL`, are set, Mono will use local. See mono/utils/mono-dl-fallback.h for more info.
+
+This also means that dynamically linking libmonosgen and attempting to resolve Mono symbols from `dlopen(NULL, ...)` will no longer work. `__Internal` has been preserved as a Mono-specific extension, but its meaning has been expanded. When P/Invoking into `__Internal`, the runtime will check both `dlopen(NULL)` and the runtime library in the case that they differ, so that users attempting to call Mono APIs with `__Internal` will not have those calls break.
+
+Mono now supports the `DefaultDllImportSearchPathsAttribute` attribute! In particular, passing `DllImportSearchPath.AssemblyDirectory` is now required to have the loader search the executing assembly's directory for native libraries.
```
lambdageek: This paragraph (and the next couple) are separate changes, right?

It might be better to give this article some more structure (ie either add some headings, or make the whole thing a big `<ul>`)

---

Or maybe, since the other points are short, add the summary list at the beginning that just says "changed default from RTLD_GLOBAL to RTLD_LOCAL, see below for ramifications and caveats"  for the big change and add the smalelr ones just to the summary.

lambdageek (commented):
Had an idea of how to structure this a bit clearer

_posts/2020-08-24-native-loader-net5.md
CoffeeFlux: It does different things on Linux and MacOS, so I'd rather not get too into it. I'll link to man pages and note the difference.

_posts/2020-08-24-native-loader-net5.md
CoffeeFlux: I like the latter option, will update.

_posts/2020-08-24-native-loader-net5.md
CoffeeFlux: Oh yeah, oops. Works for me.

_posts/2020-08-24-native-loader-net5.md
CoffeeFlux: @lambdageek took another swing at this, let me know if you think this is clearer.

lambdageek (approved):

CoffeeFlux:
Oops, almost forgot about this. @marek-safar I believe you wanted to see this post before it went live? If you're fine with it, I'll merge.

marek-safar:
Nice! I'm wondering if https://github.com/dotnet/designs would be a better place for this

CoffeeFlux:
I wasn't sure. It's mostly a list of things you might care about when migrating to net5 rather than a real design doc, so I thought a blog post was more appropriate? Happy to put it wherever you think it's most useful.

marek-safar:
dotnet/designs is not for design docs only. Happy to land this as a blog post first

CoffeeFlux:
All I see in the dotnet/designs issues are design/API proposals, so if its scope is intended to extend beyond that it would be nice to see something explaining the intended purpose.

I think eventually we'll want a larger "migration a framework Mono app to .NET 6" document once we have the embedding API changes pinned down, and maybe that would be more appropriate there? I'm happy to spearhead that effort when the time comes.

CoffeeFlux:
I guess short-term we could do a more complete "migrating to net5" one as well, even though we haven't figured out the embedding API yet? Up to you.

PR #523, opened 2020-09-30
Added Doctype html

Spiritman021:


akoeplinger:
This file is a template that gets processed by Jekyll so it doesn't need a doctype.

PR #536, opened 2020-10-01
Improve Docs

nowitsbalibhadra:


jarena3:
@nowitsbalibhadra 
do you really need a t-shirt that badly? fuckin loser

PR #538, opened 2020-10-01
Update 404.html

kanubhav906:
Added more informative error message for user convenience.

### DIFF
--- a/404.html
+++ b/404.html
@@ -6,8 +6,13 @@ permalink: /404.html
 <div class="row">
   <div class="small-12 columns">
      <h1>We could not find that page</h1>
-     <article>
-     Please check that the link you used is correct.
-     </article>
+     <p>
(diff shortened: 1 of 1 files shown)


404.html
```diff
@@ -6,8 +6,13 @@
...
      <h1>We could not find that page</h1>
-     <article>
-     Please check that the link you used is correct.
-     </article>
+     <p>
+     <h3>If that address is correct, here are three other things you can try:</h3>
+     <ul>
+       <li>Try again later.</li>
+       <li>Check your network connection.</li>
+       <li>If you are connected but behind a firewall, check that your browser has permission to access the Web.</li>
```
akoeplinger, line 14: Thank you, but none of these will actually help the user since if they can see our 404 page they have a working internet and network connection and trying again later won't make the page appear :)

PR #548, opened 2021-02-01
Clean up x86 page

squ1dd13:
I've made a few formatting changes, but I've also removed the statement "Instruction selector is able to pick best instructions on x86". I don't think it makes sense as an "optimization" given that it should be a given that the compiler chooses instructions from the correct instruction set. I could have misinterpreted that though, so I'm happy to add it back if need be.

### DIFF
--- a/docs/about-mono/supported-platforms/x86.md
+++ b/docs/about-mono/supported-platforms/x86.md
@@ -1,23 +1,20 @@
 ---
-title: X86
+title: x86
 redirect_from:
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/mono/website?pullRequest=548) <br/>All CLA requirements met.

akoeplinger (approved):
Thank you!

PR #550, opened 2021-04-15
Fixed grammar and links

miraantolovich:
Fixed grammar by adding periods, changing .Net to .NET, and mono to Mono.
Removed a broken link (lead to gambling website)
Updated links to lead to correct download tab (ex: Mac now leads to Mac download)


akoeplinger (approved):
thank you!

PR #559, opened 2021-06-08
Update index.html

billycgn:
Redirect to your Website if user visits the Github Page.


akoeplinger:
We already have that code in https://github.com/mono/website/blob/821450905be2c52fe5727af4aa23c7dd77271649/_layouts/base.html#L24-L27 so this shouldn't be needed.

PR #576, opened 2021-12-22
Migrate apt scripts to use Signed-By feature

andrediogo92:
Changes apt-key usages to saving a local keyring directly via gpg. No new dependencies are introduced. Updates all apt source list configurations (Ubuntu, Debian and Raspbian) to use the signed-by feature pointing to this local keyring.

Usage of apt-key will be deprecated as per the manual apt-key(8): "apt-key will last be available in Debian 11 and Ubuntu 22.04."

Usage of global trusted PGP keys is also deprecated as they are widely insecure: they are unconditionally trusted by APT on all other repositories configured on the system that don't have a signed-by option, even the official Debian / Ubuntu repositories. As a result, any unofficial APT repository which has its signing key added to /etc/apt/trusted.gpg or /etc/apt/trusted.gpg.d can replace any package on the system.

According to the Debian wiki, the more proper and secure way of adding third-party repositories is to use the Signed-By feature in the apt sources list configuration. This should point to a local keyring, usually defined in /usr/share/keyrings/, with a trailing 'archive-keyring' postfix.

See: https://wiki.debian.org/DebianRepository/UseThirdParty
See: https://michael-prokop.at/blog/2021/02/16/how-to-properly-use-3rd-party-debian-repository-signing-keys-with-apt/

### DIFF
--- a/_includes/download.html
+++ b/_includes/download.html
@@ -94,42 +94,42 @@
             <p>The package repository hosts the packages you need, add it with the following commands.</p>
             <p><i>Note:</i> the packages should work on newer Ubuntu versions too but we only test the ones listed below.</p>
             <h3>Ubuntu 20.04 <small>(amd64, armhf, arm64, ppc64el)</small></h3>
-{% if include.releasename == "Nightly" %}
-<pre><code class="language-bash">sudo apt install gnupg ca-certificates
-sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
-echo "deb https://download.mono-project.com/repo/ubuntu {{ include.releasename | downcase }}-focal main" | sudo tee /etc/apt/sources.list.d/mono-official-{{ include.releasename | downcase }}.list
-echo "deb https://download.mono-project.com/repo/ubuntu preview-focal main" | sudo tee /etc/apt/sources.list.d/mono-official-preview.list
+            {% if include.releasename == "Nightly" %}
+<pre><code class="language-bash">sudo apt install apt-transport-https ca-certificates gnupg
+sudo gpg --no-default-keyring --keyring /usr/share/keyrings/mono-official-archive-keyring.gpg --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
+echo "deb [signed-by=/usr/share/keyrings/mono-official-archive-keyring.gpg] https://download.mono-project.com/repo/ubuntu {{ include.releasename | downcase }}-focal main" | sudo tee /etc/apt/sources.list.d/mono-official-{{ include.releasename | downcase }}.list
+echo "deb [signed-by=/usr/share/keyrings/mono-official-archive-keyring.gpg] https://download.mono-project.com/repo/ubuntu preview-focal main" | sudo tee /etc/apt/sources.list.d/mono-official-preview.list
 sudo apt update</code></pre>
 {% else %}
-<pre><code class="language-bash">sudo apt install gnupg ca-certificates
-sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
-echo "deb https://download.mono-project.com/repo/ubuntu {{ include.releasename | downcase }}-focal main" | sudo tee /etc/apt/sources.list.d/mono-official-{{ include.releasename | downcase }}.list
+<pre><code class="language-bash">sudo apt install apt-transport-https ca-certificates gnupg
+sudo gpg --no-default-keyring --keyring /usr/share/keyrings/mono-official-archive-keyring.gpg --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
+echo "deb [signed-by=/usr/share/keyrings/mono-official-archive-keyring.gpg] https://download.mono-project.com/repo/ubuntu {{ include.releasename | downcase }}-focal main" | sudo tee /etc/apt/sources.list.d/mono-official-{{ include.releasename | downcase }}.list
 sudo apt update</code></pre>
 {% endif %}
             <h3>Ubuntu 18.04 <small>(i386, amd64, armhf, arm64, ppc64el)</small></h3>
 {% if include.releasename == "Nightly" %}
-<pre><code class="language-bash">sudo apt install gnupg ca-certificates
-sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
-echo "deb https://download.mono-project.com/repo/ubuntu {{ include.releasename | downcase }}-bionic main" | sudo tee /etc/apt/sources.list.d/mono-official-{{ include.releasename | downcase }}.list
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/mono/website?pullRequest=576) <br/>All CLA requirements met.

_includes/download.html
```diff
@@ -94,42 +94,48 @@ <h2>
...
             <p><i>Note:</i> the packages should work on newer Ubuntu versions too but we only test the ones listed below.</p>
             <h3>Ubuntu 20.04 <small>(amd64, armhf, arm64, ppc64el)</small></h3>
-{% if include.releasename == "Nightly" %}
-<pre><code class="language-bash">sudo apt install gnupg ca-certificates
-sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
-echo "deb https://download.mono-project.com/repo/ubuntu {{ include.releasename | downcase }}-focal main" | sudo tee /etc/apt/sources.list.d/mono-official-{{ include.releasename | downcase }}.list
-echo "deb https://download.mono-project.com/repo/ubuntu preview-focal main" | sudo tee /etc/apt/sources.list.d/mono-official-preview.list
+            {% if include.releasename == "Nightly" %}
+<pre><code class="language-bash">sudo apt install apt-transport-https ca-certificates gnupg
+sudo mkdir -m 700 /root/.gnupg
```
akoeplinger: why is this necessary?

_includes/download.html
```diff
@@ -94,42 +94,48 @@ <h2>
...
-<pre><code class="language-bash">sudo apt install gnupg ca-certificates
-sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
-echo "deb https://download.mono-project.com/repo/ubuntu {{ include.releasename | downcase }}-bionic main" | sudo tee /etc/apt/sources.list.d/mono-official-{{ include.releasename | downcase }}.list
+<pre><code class="language-bash">sudo apt install apt-transport-https ca-certificates gnupg
+sudo mkdir -m 700 /root/.gnupg
+sudo gpg --no-default-keyring --keyring /usr/share/keyrings/mono-official-archive-keyring.gpg --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
+echo "deb [signed-by=/usr/share/keyrings/mono-official-archive-keyring.gpg] https://download.mono-project.com/repo/ubuntu {{ include.releasename | downcase }}-bionic main" | sudo tee /etc/apt/sources.list.d/mono-official-{{ include.releasename | downcase }}.list
 sudo apt update</code></pre>
 {% endif %}
             <h3>Ubuntu 16.04 <small>(i386, amd64, armhf, arm64, ppc64el)</small></h3>
```
akoeplinger, line 122: did you check that this works on Ubuntu 16.04? reading the blog you linked there are some apt features required that might not be in 16.04

_includes/download.html
```diff
@@ -94,42 +94,48 @@ <h2>
             <p>The package repository hosts the packages you need, add it with the following commands.</p>
             <p><i>Note:</i> the packages should work on newer Ubuntu versions too but we only test the ones listed below.</p>
             <h3>Ubuntu 20.04 <small>(amd64, armhf, arm64, ppc64el)</small></h3>
-{% if include.releasename == "Nightly" %}
-<pre><code class="language-bash">sudo apt install gnupg ca-certificates
-sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
-echo "deb https://download.mono-project.com/repo/ubuntu {{ include.releasename | downcase }}-focal main" | sudo tee /etc/apt/sources.list.d/mono-official-{{ include.releasename | downcase }}.list
-echo "deb https://download.mono-project.com/repo/ubuntu preview-focal main" | sudo tee /etc/apt/sources.list.d/mono-official-preview.list
+            {% if include.releasename == "Nightly" %}
+<pre><code class="language-bash">sudo apt install apt-transport-https ca-certificates gnupg
```
akoeplinger: we don't need to install `apt-transport-https` on Ubuntu 18.04 and later since apt supports https directly there

akoeplinger:
@Seriyin thanks for the PR and sorry for the delay, I left a few comments but otherwise it looks good 😄

_includes/download.html
andrediogo92: Unfortunately writing to /usr/share/keyrings requires privileged access.
Running gpg with sudo makes the user root which then requires accessing root's home folder to cache temporary keyring before moving it to its final destination.
This usage of gpg is not ideal and while unlikely to ever be deprecated requires this ugly hack which, agreed, I am also not very fond of.

_includes/download.html
andrediogo92: By my searching, 16.04 seems to ship apt 1.2.35 which definitely does not have support for armored gpg keys but if I recall correctly that is not the case for the mono gpg key.
It should support signed-by feature just fine since it shipped with apt 1.1.

I did not run it on 16.04 yet but might try it in a bit.

andrediogo92:
As a possible different usage scenario, hosting the gpg key statically on one of mono's repos might be easier to support.

That way the public key could just be curl'd by the user directly bypassing this particular edge-case usage of gpg.

I haven't checked yet if its already hosted in any of them directly since all instructionals I've seen used keyservers to retrieve the key.

dogsbody:
I came here as for the last year we've been installing Ubuntu 20.04 instead of Ubuntu 22.04 purely because of this bug. This weekend I was determined to fix it and was going to make a pull request only to find this one already here for 16 months :-( 

@andrediogo92 is correct, it seems the keys already exist (in two forms) on the download server. Instead of messing around with the Ubuntu keyserver and gpg home directory permissions, either of these other two commands will work...

    curl -sS https://download.mono-project.com/repo/xamarin_ring.gpg | sudo tee /usr/share/keyrings/mono-official-archive-keyring.gpg >/dev/null

    curl -sS https://download.mono-project.com/repo/xamarin.gpg | gpg --dearmor | sudo tee /usr/share/keyrings/mono-official-archive-keyring.gpg >/dev/null

The repo lines look perfect.  

How can I help get this into production? Is it worth trying to update this pull request or make a new one?

akoeplinger:
@dogsbody sorry this dropped off my radar since I wanted to see whether there's some alternative to messing with the gpg home directory permissions. As far as I know the reason why we're using keyservers is to provide defense-in-depth in case the download server is compromised so you can't tamper with both the key and packages.

After googling again I just found a way to avoid the permission issue though in https://www.digitalocean.com/community/tutorials/how-to-handle-apt-key-and-add-apt-repository-deprecation-using-gpg-to-add-external-repositories-on-ubuntu-22-04#addendum-adding-an-external-repository-using-a-keyserver, apparently `gpg` supports a `--homedir` argument so I'll update the PR to use that instead :)

_includes/download.html
akoeplinger: I tried on Ubuntu 16.04 and it doesn't seem to work there even though it has apt 1.2.35:

```
W: GPG error: https://download.mono-project.com/repo/ubuntu stable-xenial InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY A6A19B38D3D831EF
E: The repository 'https://download.mono-project.com/repo/ubuntu stable-xenial InRelease' is not signed.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

I reverted those instructions back to the original.

akoeplinger (approved):

PR #640, opened 2024-08-20
Update banshee link

popey:
The old domain has long lapsed. The link I provided is the best target for this entry, I think. 🙏


akoeplinger (approved):

popey:
@dotnet-policy-service agree

PR #667, opened 2026-01-27
Add .net 4.8 to the compatibility table

LazaroOnline:
Since mono 6.6.0 it supports .net 4.8 (but not 4.8.1). 
https://www.mono-project.com/docs/about-mono/releases/6.6.0/#net-48-reference-assemblies

And probably newer versions of C# too.


LazaroOnline:
@dotnet-policy-service agree

akoeplinger (approved):

